PR #9, opened 2016-08-01
Large enclave support

wh0:


andyzyb:
Current SGX hardware has limited EPC size and it cannot support big Enclave.  We won't be able to load a really big Enclave to fully test the patch. 

We will re-evaluate the patches when it comes to the point that supporting 64-bit size Enclave is real.  There might be other places we need to fix in addition to these. 

Thanks for looking at the code and providing patches.

wh0:
It doesn't have to be really big. You could test with an enclave barely bigger than 4GB, couldn't you? I was able to run a 9.1GB (0x220000000) enclave this way.

The Linux SGX driver supports paging EPC to main memory, doesn't it?

Even if there is a low limit, this patch prevents large numbers in the XML, such as 0x100100000 from being silently truncated and compiled as 0x100000, instead of, say, raising an error.

andyzyb:
Good to know the indication that it runs with the patch for big Enclave.  How much physical memory on your system to run 9.1 GB Enclave?  We'll see if that can be reproduced.

wh0:
16 GB

sean-jc:
+1  The XML changes are necessary to support exposing XFRM/XFRM_MASK to the user.

PR #10, opened 2016-08-10
Update 'echo' to 'echo -e' in psw/ae/buildenv.mk

pgonda:
Without the '-e' echo fails to escape the '\n' on bash version 4.3.11


andyzyb:
Please look at the contributing.MD, and follow the process.  A sign-off is needed for pull request.

zhaohuidu:
My test shows 'echo -e' doesn't work under ubuntu 14.04.
The google result shows that by default Makfile uses dash instead of bash where the parameter list of echo is different from that in bash

PR #17, opened 2016-09-20
Add libunwind generated files dir to .gitignore

sean-jc:


yuyuany:
We may do minor changes to enable libunwind within enclave. Not sure if we need add files to the libunwind directory in the future. We'd better not add it to .gitignore.

sean-jc:
Argh, sorry, saw `src` in the untracked files and assumed the entire directory was generated.

PR #18, opened 2016-09-20
Skip EEXTEND for .bss sections

sean-jc:
.bss sections reserve space for uninitialized data.  Applications cannot
rely on uninitialized data to hold a specific value, thus measuring .bss
sections provides no value or additional security.

This change significantly speeds up `sgx_create_enclave` for applications with large .bss sections (large buffers, custom memory allocation, etc...), e.g. create time with a 12mb .bss is ~20ms versus ~100ms without the change.

**WARNING:** This change breaks backwards compatibility, i.e. it requires prebuilt enclaves to be rebuilt.  A version which allows applications to opt-in to the behavior can be found [here](https://github.com/sean-jc/linux-sgx/tree/bss_opt_in_no_eextend), which allows utilizing this behavior while still being able to load existing prebuilt enclaves.

### DIFF
--- a/psw/urts/loader.cpp
+++ b/psw/urts/loader.cpp
@@ -171,7 +171,7 @@ int CLoader::build_mem_region(const section_info_t * const sec_info)
     {
         size_t size = (size_t)(ROUND_TO_PAGE(sec_info->virtual_size) - ROUND_TO_PAGE(sec_info->raw_data_size));
         sinfo.flags = sec_info->flag;
-        if(SGX_SUCCESS != (ret = build_pages(rva, size, 0, sinfo, ADD_EXTEND_PAGE)))
+        if(SGX_SUCCESS != (ret = build_pages(rva, size, 0, sinfo, ADD_PAGE_ONLY)))
             return ret;
     }
 


zhaohuidu:
The initial value of all global symbols inside .bss section must be zeroed. So EXTENTED is required for .bss section.

sean-jc:
Bummer, didn't realize zeroing the bytes was explicitly stated in language specs.  

Would you be open to merging a patch that allows a user to define a section that is not measured?  This would allow applications to declare large buffers or memory pools without having to play tricks with the heap or take the EEXTEND performance hit.

For example, add this macro to sgx_defs.h and then skip EEXTEND for the .sgx.unmeasured section.

```c
#define __unmeasured __attribute__((__section__(".sgx.unmeasured, \"wa\", @nobits#")))
```

yuyuany:
Zero the bytes may not be required by the language spec, but as an enclave, we needs each page added to it is predefined. If a page that is not measured is exploited and inserted some shell code, we do not know what will happen. 

In the latest design, only heap is not measured, and it is zeroed by tRTS during the enclave initialization. But I don't think all the sections could do such operations. tRTS may write some data in .bss section during the enclave initialization, so we cannot clear it. 

Enclave loading time is always the critical performance data. Your proposal is really valuable and we may consider in the future release.

sean-jc:
Closing, will file a new PR if I get an opt-in model working.

PR #19, opened 2016-09-22
Update .gitignore files to ignore known artifacts

sean-jc:
Ignore output and generated files from libunwind, rdrand, installer and
AESM.  Ignore all known executables that are created, e.g. sgx_sign.

Signed-off-by: Sean Christopherson sean.j.christopherson@intel.com

I split out libunwind and rdrand into separate files to keep the top-level .gitignore cleaner, let me know if you would prefer to have everything in a single file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@
 *.so
 *.so.*
 
+#map files
+*.map
+
 #generated proxy
 *_u.c
 *_u.h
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -8,12 +8,37 @@
...
+*.map
+
 #generated proxy
 *_u.c
 *_u.h
 *_t.c
 *_t.h
 
 #output files
+output/
```
andyzyb: Adding file/folder paths in .gitignore doesn't seem an extensible approach. We may change the layout of the generated files, folders where temp files are put, etc.   Then it will require to change .gitinore file as well.  

We'd like to just have a simple rule for .gitinore which just ignore some specific file (with specific extensions), but not specify the path.

andyzyb:
No going to accept the patch to list all the generated files in .gitignore in short term.

PR #22, opened 2016-09-27
Clean up addition of guard pages between sections

sean-jc:
Sort sections by ascending rva before processing them in build_sections.
Insert guard pages between all sections if there is at least one unused
page between sections.  Unconditionally add a guard page after the last
section.  Do not use the previous section's si_flags for the permissions
of each guard page, guard pages should not be accessible, i.e. si_flags
should be '0'.

EDIT: Updated to match the commit message.  Modified the PR to keep the original behavior of adding a final guard page in build_sections.  From a security perspective, build_sections should ensure a final guard page after its last section.

<br>
This patch assumes the intent of the `last_section` code in build_sections is to add guard pages whenever possible.  If the intent is something else entirely then this patch is probably breaking something.

The current code is inconsistent with regard to adding guard pages between sections/segments.  The algorithm for determining if a guard page should be added is heavily dependent on the size of the last section and results in what is essentially arbitrary application of guard pages.

```c
ROUND_TO_PAGE(last_section->virtual_size() + last_section->get_rva()) < ROUND_TO_PAGE(ROUND_TO_PAGE(last_section->virtual_size()) + last_section->get_rva())
```

The table below highlights the inconsistencies; two sections will have different guard page behavior even though the memory layout of the sections is nearly identical.  `End RVA` is the inclusive last byte of the section, `Guarded` reflects whether or not a guard page will be added, and `Buffer` is the minimum number of unused bytes between the last section and the next used page.  In the extremes, `Buffer` can vary from zero bytes to nearly two full pages.

| RVA    | Size   | End RVA  | Guarded | Buffer |
| ------ | ------ | -------- | ------- | ------ |
| 0x0400 | 0x03FF | 0x07FE   | Yes     | 0x0C02 |
| 0x0400 | 0x0400 | 0x07FF   | Yes     | 0x0C01 |
| 0x0400 | 0x0401 | 0x0800   | Yes     | 0x1C00 |
| 0x0FFF | 0x0400 | 0x13FE   | No      | 0x0C02 |
| 0x1000 | 0x0400 | 0x13FF   | No      | 0x0C01 |
| 0x1001 | 0x0400 | 0x1400   | No      | 0x0C00 |
| 0x0400 | 0x0FFF | 0x13FE   | No      | 0x0C02 |
| 0x0400 | 0x1000 | 0x13FF   | No      | 0x0C01 |
| 0x0400 | 0x1001 | 0x1400   | Yes     | 0x1C00 |
| 0x0FFF | 0x0400 | 0x13FE   | No      | 0x0C02 |
| 0x1000 | 0x0400 | 0x13FF   | No      | 0x0C02 |
| 0x1001 | 0x0400 | 0x1400   | No      | 0x0C02 |
| 0x1000 | 0x0FFF | 0x1FFE   | No      | 0x0001 |
| 0x1000 | 0x1000 | 0x1FFF   | No      | 0x0000 |
| 0x1000 | 0x1001 | 0x2000   | Yes     | 0x1FFF |
| 0x1001 | 0x0FFF | 0x1FFF   | Yes     | 0x1000 |
| 0x1001 | 0x1000 | 0x2000   | No      | 0x0FFF |
| 0x1001 | 0x1001 | 0x2001   | Yes     | 0x1FFE |

### DIFF
--- a/psw/urts/loader.cpp
+++ b/psw/urts/loader.cpp
@@ -182,32 +182,41 @@ int CLoader::build_sections(vector<uint8_t> *bitmap)
 {
     int ret = SGX_SUCCESS;
     std::vector<Section*> sections = m_parser.get_sections();
-    uint64_t max_rva =0;
     Section* last_section = NULL;
+
+    // Return now if there are no sections to process.
+    if (sections.empty())
+        return SGX_SUCCESS;
+
+    // Sort the sections by ascending rva, this allows for the easy insertion of
+    // guard pages between sections.
+    sort(sections.begin(), sections.end(),
+        [](const Section * a, const Section * b) -> bool
+    {
+        return a->get_rva() < b->get_rva();
+    });
 
     for(unsigned int i = 0; i < sections.size() ; i++)
     {
-        
-        
-        
+        // If this is not the first section, attempt to insert a guard page at
+        // the end of the last section.  A guard page can be inserted if there
+        // is at least one unused page between the last section and this section.
+        // Sections are sorted by ascending address, so we don't need to do any
+        // fancy tracking to detect unused pages, simply compare the end of the
+        // last section (rounded up) to the start of this section (rounded down).
         if((last_section != NULL) &&
-           (ROUND_TO_PAGE(last_section->virtual_size() + last_section->get_rva()) < ROUND_TO_PAGE(ROUND_TO_PAGE(last_section->virtual_size()) + last_section->get_rva())) &&
-           (ROUND_TO_PAGE(last_section->get_rva() + last_section->virtual_size()) < (sections[i]->get_rva() & (~(SE_PAGE_SIZE - 1)))))
+           (ROUND_TO_PAGE(last_section->get_rva() + last_section->virtual_size()) < TRIM_TO_PAGE(sections[i]->get_rva())))
         {
-            size_t size = SE_PAGE_SIZE;
+            // sinfo.flags is intentionally left as '0', the guard page should
+            // not be accessible, i.e. not readable, writable or executable.
             sec_info_t sinfo;
             memset(&sinfo, 0, sizeof(sinfo));
-            sinfo.flags = last_section->get_si_flags();
             uint64_t rva = ROUND_TO_PAGE(last_section->get_rva() + last_section->virtual_size());
-            if(SGX_SUCCESS != (ret = build_pages(rva, size, 0, sinfo, ADD_EXTEND_PAGE)))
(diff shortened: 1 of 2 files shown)


andyzyb:
The first thing we will need to think to change the logic of Enclave loading is the backward compatibility.  Will this patch break the backward compatibility that if the enclave is built and signed with old sign tool, then it will fail to load with the this patch committed?

sean-jc:
This patch does cause a change in an enclave's measurement, i.e. breaks backwards compatibility in that sense.

That being said, this patch isn't a change in the design logic, but rather it is a bug fix for a security hole.  In the current implementation, the addition of a guard page is not guaranteed even when there is sufficient space in the enclave to insert a guard page.  The presence of a guard page can change between enclave builds based purely on the size of the enclave's section, e.g. removing a line of code could cause a guard page to (dis)appear.

mikehearn:
Given how early SGX on Linux is, I would not be too afraid of backwards compatibility breaks at this point?

andyzyb:
The better way to fix the issue is to introduce a new version of loader which can load both old Enclaves and new Enclaves. We are going to look into it.   The backward compatibility is important because some pre-built Enclaves are out there.

sean-jc:
>The better way to fix the issue is to introduce a new version of loader which can load both old Enclaves and new Enclaves.

What about providing the option to embed the loader in their application?  The architectural enclaves could still be loaded from a centralized daemon, but individual applications could have a loader specific to their enclave.  I can see this being very desirable from a deployment perspective as it would allow users to change the loader behavior for their application without having to push a new AESM or coordinate upgrades between multiple applications.

ZongminGu:
The "guard page" guarantee the enclave gets an exception, if any code touch the page. The current implementation is leaving the "guard page" area empty. The whole enclave space is measured by CPU, so it is good solution, work, high loading performance and small EPC footprint.

Add a no R/W/X page also should work, but lost the benefits.

sean-jc:
>The current implementation is leaving the "guard page" area empty.

This is true for the guard pages inserted into the context, e.g. between heap, stack, TCS, etc..., but the additional pages being inserted in `build_sections` are being added with the flags of the last section.  My assumption from reading the code is that there was a desire to explicitly add a trailing blank page after sections, hence this patch to ensure said page is added when possible and is not accessible.  If the intent is to skip unused pages, à la the context's guard pages, then the code in question should simply be deleted.  If the intent is something else entirely then I think comments detailing the intent of the code are needed, because it is not at all clear what the code is supposed to be doing or what value it adds.

ZongminGu:
You are right, the enclave add a black page at the tail in some cases. But it is not a guard page. Actually, it is compatible fix to support existing enclaves build with old SDK. In the old SDK we have a bug to add such a page at the tail. The SDK add such page for backward compatibility.

PR #23, opened 2016-09-28
Add support for declaring unmeasured data

sean-jc:
Add recognition of a special ELF section, '.sgx.unmeasured', that can
be used by enclaves to declare data that should not contribute to the
measurement of an enclave.  The section's address and size must both
be page aligned.  Pages that are wholly contained by the unmeasured
sections are built with ADD_PAGE_ONLY, i.e. they skip EEXTEND.

Report an error if the unmeasured section is not properly aligned, or
if the section contains any initialized data.  Assert if the section
straddles multiple segments (invalid ELF file).

Define the macros '__unmeasured' and '__unmeasured_aligned' for use
by enclave developers in declaring unmeasured data.  The aligned macro
is intended to be used once per enclave to ensure proper alignment.

Signed-off-by: Sean Christopherson sean.j.christopherson@intel.com

<br>
**Summary**
This is essentially a RFC on the concept of providing enclave developers with the option of declaring data as unmeasured.  

On my system, EEXTEND adds a latency of 25-42 microseconds per 4k page, e.g. an application with 12MiB of unmeasured data saves ~80 milliseconds during enclave creation when compiled with this patch.  If the majority of the code/data being loaded for the application is unmeasured data, then this can result in a 2x-5x reduction in time spent creating the enclave.

The intended usage model of __unmeasured is for cases where the data is guaranteed to be accessed in a read-after-write model, e.g. buffers, static memory pools, etc...  Without defined workloads it's difficult to determine how many users would benefit from this feature, but I think there is value in being able to "allocate" large chunks of memory without having to suffer the overhead of EEXTEND or coordinate (and fine tune) the heap size with the code base.

<br>
**Sample Usage:**

``` c
static STREAM_TYPE __unmeasured_aligned a[STREAM_ARRAY_SIZE+OFFSET], __unmeasured b[STREAM_ARRAY_SIZE+OFFSET], __unmeasured c[STREAM_ARRAY_SIZE+OFFSET];
```

<br>
**Results:**
The table below shows the latency, in seconds, of sgx_create_enclave for an enclave that defines a large buffer using each of the methods.  Allocations other than the buffer are minimal, e.g. 16kb of stack and ~80kb of code/data.  For any sizable buffer, skipping EEXTEND for the buffer via the heap or .sgx.unmeasured yields significantly lower latency during enclave creation.

I'm in the process of investigating why using the heap is slower than .sgx.unmeasured, I expected them to be roughly equivalent since this benchmark does not incorporate any actual malloc calls.

| Method | 72mb | 12mb | 384kb |
| --- | --- | --- | --- |
| .bss | 0.594111 | 0.101852 | 0.004779 |
| heap | 0.147291 | 0.026281 | 0.002564 |
| .sgx... | 0.104936 | 0.019890 | 0.002129 |

### DIFF
--- a/common/inc/sgx_defs.h
+++ b/common/inc/sgx_defs.h
@@ -51,4 +51,7 @@
 
 #define SGX_NOCONVENTION /* Empty.  No calling convention specified. */
 
+#define __unmeasured __attribute__((__section__(".sgx.unmeasured, \"wa\", @nobits#")))
+#define __unmeasured_aligned __unmeasured __attribute__((aligned(0x1000)))
+
 #endif /* !_SGX_DEFS_H_ */
--- a/psw/urts/loader.cpp
+++ b/psw/urts/loader.cpp
@@ -171,6 +171,34 @@ int CLoader::build_mem_region(const section_info_t * const sec_info)
     {
         size_t size = (size_t)(ROUND_TO_PAGE(sec_info->virtual_size) - ROUND_TO_PAGE(sec_info->raw_data_size));
         sinfo.flags = sec_info->flag;
+
+        // Pages that contain **only** uninitialized data may be part of the
+        // .sgx.unmeasured section, in which case they build with ADD_PAGE_ONLY.
+        // If the application defined an unmeasured section and this section
+        // overlaps the unmeasured section, then we need to split this section
+        // and make multiple calls to build_pages.
+        uint64_t unmeasured_size = m_parser.get_unmeasured_size();
+        uint64_t unmeasured_start = m_parser.get_unmeasured_address();
+        uint64_t unmeasured_end = unmeasured_start + unmeasured_size;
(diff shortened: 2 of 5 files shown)


sean-jc:
The performance delta between `heap` and `.sgx.unmeasured` is due to `do_init_enclave` calling `memset` to zero out the heap, while the unmeasured data is not zeroed out during creation.

andyzyb:
The design of measuring all the pages except heap (but heap is zeroed out during Enclave initialization) is a defense in depth to avoid potential unmeasured memory being utilized by attackers.  Adding more unmeasured area probably makes it easier to be used incorrectly and introduces security bugs.  

The performance improvement is an important aspect, but this new feature will need to go through some security review before adopting.

andyzyb:
no longer a valid PR.

PR #41, opened 2016-11-07
Bug-fix #40

### LINKED ISSUE #40: Bug in build-installpkg.sh script
The function _get_arch()_ from this script expects the output of the _readelf_ command to contain the string "Class:" along with the information needed. This will fail on an OS configured in other languages other than English, causing an error later on, since the installers will be using the wrong architecture (x86 instead of x64).

This bug can be reproduced in the latest master. (Running on a Dell Optiplex 5040 with i7-6700, SGX is enabled, Ubuntu 14.04.5 (64 bit), PT_BR system language).

rodolfoams:
Changed get_arch() function to work independently on the OS's language.


andyzyb:
Please look at the contributing.md and make sure you accept the license agreement, and each patch needs a signed-off-by before it can be accepted.

rodolfoams:
Added a signed-off-by to each patch. I accept the license agreement.

PR #50, opened 2016-12-03
Fixes for various compilation issues

jsakkine-intel:


lzha101:
For commit aee3e99 'Set memory protections before building the enclave',  the set_memory_protection() function should not be called before loading enclave but after loading enclave. Setting memory protections before enclave loading may break enclave loading process. We have some SGX specific sections such as TCS which need to be loaded with specific attributes. And we may also have some text sections with TEXTRELs which need to be loaded with extra write attribute.  After the enclave loading, the set_memory_protection() API is called to change these section with regular attributes.

jsakkine-intel:
I see your point. No matter how you implement the relocator you have to do it after EINIT so you always need to do mprotect calls also after EINIT, am I correct? If I ease the driver requirement to not allow VMAs to be closed that should also work for you and not break anything? This will still allow you to do mprotect calls whenever you want to.

Please read this email message I wrote: https://lists.01.org/pipermail/intel-sgx-kernel-dev/2016-December/000195.html

Thanks.

jsakkine-intel:
that commit is no longer part of the pull request

jsakkine-intel:
Hi, I also dropped the commit changing the download script.

kaimast:
Aww. Please merge this. Would have saved me half an hour of googling that weird protobuf issue.

jsakkine-intel:
Don't know why it wasn't merged and now there's some other stuff that causes merge conflict :(

jsakkine-intel:
Rebased.

lzha101:
Thanks for the rebase. The  pull request was planned to be merged after the 1.7 release, because the request was created when we were working on the 1.7 release.   We will merge it as soon as possible.

linux/installer/bin/install-sgx-psw.bin.tmpl
```diff
@@ -30,18 +30,6 @@
 #
 #
 
-modinfo isgx &> /dev/null
-if [[ $? != "0" ]]; then
-    lsmod | grep isgx &> /dev/null
-    if [[ $? = "0" ]]; then
-        echo "Warning: You did not follow the document to install the driver package"
```
lzha101: Based on current code/product status,  the driver dependence check is still valid and it would help SGX newbies, so we decide to keep the check in the PSW installation script.   If the driver name changes in future, only changing the PSW installation script is not enough. We also need to change the loader implementation. 
@jsakkine-intel, could you help to drop this commit? Then we can merge this pull request.

jsakkine-intel:
I dropped that commit.

jsakkine-intel:
Thank you.

PR #55, opened 2016-12-25
Fix building on GCC6

kaimast:
Fixed an error caused by protocol buffers and non-standard C++ code.


buildenv.mk
```diff
@@ -89,7 +89,7 @@ COMMON_FLAGS += -Wall -Wextra -Winit-self -Wpointer-arith -Wreturn-type \
 CFLAGS += -Wjump-misses-init -Wstrict-prototypes -Wunsuffixed-float-constants
 
 # additional warnings flags for C++
-CXXFLAGS += -Wnon-virtual-dtor
+CXXFLAGS += -Wnon-virtual-dtor -DPROTOBUF_INLINE_NOT_IN_HEADERS=0
```
kaimast: This is needed to prevent some warnings created by protobuf. See here for more infos https://stackoverflow.com/questions/34474091/protobuf-inline-not-in-headers#34485004

psw/uae_service/uae_wrapper/inc/AEServices.h
```diff
@@ -56,7 +56,7 @@ struct PlainData
 
     bool operator==(const PlainData& other) const {
         if (this == &other)                 return true;
-        if (this == NULL || &other == NULL) return false;
+        if (this == NULL) return false;
```
kaimast: Causes a warning by GCC. &foo can never be null.

psw/urts/tcs.cpp
```diff
@@ -150,11 +150,15 @@ int CTrustThreadPool::bind_thread(const se_thread_id_t thread_id,  CTrustThread
 
 CTrustThread * CTrustThreadPool::get_bound_thread(const se_thread_id_t thread_id)
 {
-    CTrustThread *trust_thread = NULL;
+    CTrustThread *trust_thread = nullptr;
 
-    Node<se_thread_id_t, CTrustThread*>* it = m_thread_list->Find(thread_id);
-    if(it != NULL)
-        trust_thread = it->value;
+    if(m_thread_list)
```
kaimast: You need to check for m_thread_list == nullptr here. Everything else is not standard behavior. GCC6 optimized this away and the simulator crashes... Took me a while to figure this out...

lzha101:
There is no Signed-off-by in the commits. Could you please review the CONTRIBUTING.md and add Signed-off-by in your commits? 
Some changes in this pull request are duplicated with #50, which have been applied.  And other modifications in urts will be merged after the patches have Signed-off-by.

kaimast:
I merged and added the sign-off :)

lzha101:
@kaimast, you should add the Signed-Off to every git commit message. I only see it in the commit b6d2f40.

kaimast:
See pull request @75

PR #63, opened 2017-01-24
Make heap optionally executable

exFalso:
This PR adds a new xml option `ExecutableHeap` which causes the executable flag of the enclave heap to be set.


andyzyb:
We will review this feature enhancement and provide the feedback. Thanks.

lzha101:
@exFalso, what is your purpose to make heap executable? It is not a security feature.

exFalso:
In our use case this is required for JIT compilation

MWShan:
One concern about making the heap executable relates to a future instruction support, SGX2 instructions, which will allow pages to be added to the enclave after EINIT via the EAUG instruction.  This can be used to commit a small initial  heap and then later expand the heap (the process is detailed here: [http://caslab.csl.yale.edu/workshops/hasp2016/HASP16-17.pdf](url) )
)  Any dynamically allocated heap will be committed with RW attributes because the EAUG instruction adds pages with RW permission.  The enclave will be able to change pages from RW to RX with SGX2 instructions.

exFalso:
We are trying to work with SGX1 currently, but it's great to hear about such upcoming features.

The flag is optional and is by default set to false, is this not sufficient?

andyzyb:
PR will not be merged.  Using SGX2 for enhancement is recommended.

PR #64, opened 2017-01-24
GCC 6 compatibility

exFalso:
In GCC 6 `--std=c++11` became the default, which changes the ABI. This breaks dynamic linkage with things like libprotobuf.so.8, resulting in errors like
```
messages.pb.o:(.data.rel.ro._ZTVN4aesm7message23Request_GetQuoteRequestE[_ZTVN4aesm7message23Request_GetQuoteRequestE]+0x20): undefined reference to `google::protobuf::Message::GetTypeName[abi:cxx11]() const'
(...)
% nm -D /usr/lib/libprotobuf.so.8 | grep GetTypeName | c++filt
00000000000b1b80 T google::protobuf::Message::GetTypeName() const
```

 This PR adds a flag to the build that disables the new ABI until deps are bumped to past-c++11.

### DIFF
--- a/buildenv.mk
+++ b/buildenv.mk
@@ -94,6 +94,9 @@ CXXFLAGS += -Wnon-virtual-dtor
 # for static_assert()
 CXXFLAGS += -std=c++0x
 
+# Disable cxx11 abi
+CXXFLAGS += -D_GLIBCXX_USE_CXX11_ABI=0
+
 .DEFAULT_GOAL := all
 # this turns off the RCS / SCCS implicit rules of GNU Make
 % : RCS/%,v
--- a/psw/ae/aesm_service/Makefile
+++ b/psw/ae/aesm_service/Makefile
@@ -82,6 +82,9 @@ INCLUDE += -I$(SGX_IPP_INC) \
 
 EDGER8R  := $(LINUX_SDK_DIR)/edger8r/linux/_build/Edger8r.native
 
+# Disable cxx11 abi
+CXXFLAGS += -D_GLIBCXX_USE_CXX11_ABI=0
+
 ifdef PROFILE
         CXXFLAGS += -D_PROFILE_
         CFLAGS += -D_PROFILE_
--- a/psw/uae_service/linux/Makefile
+++ b/psw/uae_service/linux/Makefile
@@ -69,6 +69,8 @@ ifeq ($(ARCH), x86)
 else
 	PROTPBUF_CXXFLAGS += -m64
 endif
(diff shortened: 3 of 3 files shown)


lzha101:
Please refer to CONTRIBUTING.md to add 'Signed-Of-by' to the commit.

Which platform are you using? And could you provide detail steps to reproduce the error?  I tried to build the code on Ubuntu14.04 with GCC6.2, and everything goes well. The protobuf library on my machine is also libprotobuf.so.8.

exFalso:
Commit signed off, thank you for pointing it out!
I'm using Arch Linux:
```
$ uname -r
4.8.13-1-ARCH
$ gcc --version | head -n 1
gcc (GCC) 6.3.1 20170109
$ pacman -Qi protobuf | grep Version
Version         : 2.5.0-4
```
Note that I had to install an old version of protobuf, as the latest is 2 major .so versions ahead (libprotobuf.so.10).

Simply running `make` should reproduce the problem.
```
$ nm ./psw/ae/aesm_service/messages.pb.o | grep GetTypeName | c++filt
                 U google::protobuf::Message::GetTypeName[abi:cxx11]() const
$ nm -D /usr/lib/libprotobuf.so.8 | grep GetTypeName | c++filt # Or wherever your protobuf is
00000000000b1b80 T google::protobuf::Message::GetTypeName() const
```
The above was run without this patch. Those symbols should line up on your system, so it's either the case you have a libprotobuf.8 with c++11 ABI, or GCC 6 on ubuntu 14.04 creates non-c++11 symbols for some reason. Could you please check by running the above on your system?

lzha101:
The GCC6 installed on my machine was built with '--with-default-libstdcxx-abi=gcc4-compatible', I suppose the GCC6 on your machine was built with '--with-default-libstdcxx-abi=new'.

lzha101:
New protobuf library may only include new CXX ABI.  If we add this option '-D_GLIBCXX_USE_CXX11_ABI=0', it will result in build failure with the new protobuf library. 
Actually if we install the dependence via the standard way (for example, via 'apt-get' on Ubuntu), there is no chance for us to encounter the incompatibility issue. 
So we decide not to merge this PR. @exFalso, you can keep the modification by yourself or you can update your protobuf, to align with your GCC version.

exFalso:
On ubuntu the latest stable (jessie) has libprotobuf.so.9, which still has non-cxx11 symbols. I think it would be safe to merge and remove the lines once it's bumped >=so.10 (that uses the new abi on ubuntu too).

lzha101:
Actually on Ubuntu16.04 (stretch), the protobuf library installed via 'apt-get' command only contains cxx11 symbols. Merging this PR will block the build on that platform.

PR #65, opened 2017-01-27
Change mallinfo field type to int as per system malloc.h

exFalso:
This PR changes the fields of `struct mallinfo` to int.

In standard malloc.h the fields of `struct mallinfo` are int, whereas in the sgx malloc.c they are defined as size_t. This causes incorrect stack allocations when calling `mallinfo()` and subsequent segfaults once the function returns(on 64bit).


andyzyb:
Please follow the sign-off-by requirement for each patch. Thanks.

lzha101:
@exFalso, we agree to accept this change. But please follow the CONTRIBUTING.md to add signed-off-by to the commit first. Thanks.

exFalso:
Hmm I already added the signed-off-by, github also shows it if you click on `...` next to the commit message.

PR #70, opened 2017-02-06
edger8r, psw: Add [enclave_id] non-pointer attribute

exFalso:
This PR adds an attribute `enclave_id` to the edl, that passes the enclave's ID to the ocall:

```
enclave {
    trusted {
        public void thello();
    };
    untrusted {
        void uhello([enclave_id] uint64_t enclave_id, [in, string] const char *name);
    };
};
```
for `uhello` will generate
tmp_u.h:
```
(...)
void SGX_UBRIDGE(SGX_NOCONVENTION, uhello, (uint64_t enclave_id, const char* name));
(...)
```
tmp_t.h:
```
(...)
sgx_status_t SGX_CDECL uhello(const char* name);
(...)
```
automatically passing the enclave id.

There may be better ways of doing this (e.g. always pass the ID), but I was trying to make the changes backwards compatible.

### DIFF
--- a/psw/urts/enclave.cpp
+++ b/psw/urts/enclave.cpp
@@ -43,7 +43,7 @@
 using namespace std;
 
 int do_ecall(const int fn, const void *ocall_table, const void *ms, CTrustThread *trust_thread);
-int do_ocall(const bridge_fn_t bridge, void *ms);
+int do_ocall(const bridge_fn_t bridge, sgx_enclave_id_t enclave_id, void *ms);
 
 CEnclave::CEnclave(CLoader &ldr)
     : m_loader(ldr)
@@ -193,7 +193,7 @@ int CEnclave::ocall(const unsigned int proc, const sgx_ocall_table_t *ocall_tabl
 
     se_rdunlock(&m_rwlock);
     bridge_fn_t bridge = reinterpret_cast<bridge_fn_t>(ocall_table->ocall[proc]);
-    error = do_ocall(bridge, ms);
+    error = do_ocall(bridge, m_enclave_id, ms);
 
     if (!se_try_rdlock(&m_rwlock))
     {
@@ -478,7 +478,7 @@ bool CEnclave::update_trust_thread_debug_flag(void* tcs_address, uint8_t debug_f
 
     if(debug_info->enclave_type == ET_DEBUG)
     {
-       
+
          if(!se_write_process_mem(pid, reinterpret_cast<unsigned char *>(tcs_address) + sizeof(uint64_t), &debug_flag2, sizeof(uint64_t), NULL))
               return FALSE;
 
@@ -490,13 +490,13 @@ bool CEnclave::update_trust_thread_debug_flag(void* tcs_address, uint8_t debug_f
 bool CEnclave::update_debug_flag(uint8_t debug_flag)
 {
     debug_tcs_info_t* tcs_list_entry = m_enclave_info.tcs_list;
-    
+
     while(tcs_list_entry)
     {
          if(!update_trust_thread_debug_flag(tcs_list_entry->TCS_address, debug_flag))
               return FALSE;
 
-         tcs_list_entry = tcs_list_entry->next_tcs_info;     
+         tcs_list_entry = tcs_list_entry->next_tcs_info;
     }
 
     return TRUE;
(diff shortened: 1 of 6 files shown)


exFalso:
I also don't know whether there is an existing way of retrieving the calling enclave's ID, but I couldn't find anything

sdk/edger8r/linux/Parser.mly
```diff
@@ -366,31 +382,34 @@ param_type: attr_block all_type {
...
-        | _ ->
-          fun is_ary ->
+          let attr = get_ptr_attr $1 in
+          if attr.Ast.pa_isptr || attr.Ast.pa_isary then fun x -> Ast.PTPtr($2, attr)
+          else
+            (* thinking about 'user_defined_type var[4]' *)
+            fun is_ary ->
             if is_ary then Ast.PTPtr($2, attr)
-            else failwithf "unexpected pointer attributes for `%s'" (Ast.get_tystr $2)
+            else failwithf "`%s' is considerred plain type but decorated with pointer attributes" s
```
mikehearn: Typo: considered

sdk/edger8r/linux/CodeGen.ml
```diff
@@ -209,8 +211,8 @@ let mk_ubridge_name (file_shortnm: string) (funcname: string) =
   sprintf "%s_%s" file_shortnm funcname
 
 let mk_ubridge_proto (file_shortnm: string) (funcname: string) =
-  sprintf "static sgx_status_t SGX_CDECL %s(void* %s)"
-          (mk_ubridge_name file_shortnm funcname) ms_ptr_name
+  sprintf "static sgx_status_t SGX_CDECL %s(sgx_enclave_id_t %s, void* %s)"
```
mikehearn: Is this new feature optional? I wonder what Intel's policy on backwards incompatible changes is.

psw/urts/enclave.cpp
```diff
@@ -43,7 +43,7 @@
 using namespace std;
 
 int do_ecall(const int fn, const void *ocall_table, const void *ms, CTrustThread *trust_thread);
-int do_ocall(const bridge_fn_t bridge, void *ms);
+int do_ocall(const bridge_fn_t bridge, sgx_enclave_id_t enclave_id, void *ms);
```
mikehearn: Is this change backwards compatible?

exFalso: It's backwards compatible in the sense that existing projects will compile fine if the files are regenerated.

sdk/edger8r/linux/Parser.mly
exFalso: Ah yes, was copypasted, but may as well fix here

ZongminGu:
Why is usage of the enclave id in the ocall?

exFalso:
> Why is usage of the enclave id in the ocall?

To know which enclave made the ocall. For example if an ocall needs to make an ecall back to the enclave, it needs to know the ID

ZongminGu:
The ecall should bind to the enclave. If there is only one enclave, the easy way is to save the enclave id as a global data after successfully loading the enclave. A better way is to create a C++ class to wrap the enclave ecalls. It should already solve the requirement "using ecall in an ocall". Is there anything special that not covered?

exFalso:
Yes, this PR is to avoid the use of a global variable which incidentally also restricts the number of enclaves to 1 per process.

Wrapping the enclave call doesn't solve anything, because the ocall table binds against static symbols, so there is no way to retrieve the wrapper object from the ocall. This could work for example if the table was bound dynamically at enclave creation with e.g. lambdas capturing the wrapper object. Wrapping the call simply changes the responsibility of passing the ID to the responsibility of passing the wrapper object pointer.

Put another way: currently ocalls are basically global functions that don't have any user-provided context. With this PR there is a way to have such a context.

(One ultra hacky way of getting the wrapper object I was thinking of would be to traverse the stack until we find the original frame that made the ecall - but of course this is insane.)

psw/urts/enclave.cpp
andyzyb: So this seems not backward compatible, because the new uRTS cannot do ocall for old Enclaves.  A backward compatible way might be uRTS knows the difference of these 2 types of ocalls and switch the ocall API during runtime.

nhynes:
related: #226

exFalso:
@nhynes @ZongminGu thank you a TLS would work of course! Closing this

PR #72, opened 2017-02-07
tlibc: Add glibc-style malloc hooks

exFalso:
This PR adds a way too hook into malloc.c functions by making the standard functions weak symbols, and exposing `__tlibc_something()` for the original functionality. This design is the same as in glibc, and allows simple hooking like so:

```
void* malloc(size_t size) {
    ocall_printf("Hello World! I'm allocating %u bytes. What a lovely day!\n", size);
    return __tlibc_malloc(size);
}
```

This has been added for `malloc`,`free`,`calloc`,`memalign` and `realloc`


ZongminGu:
If you only want to hook the malloc function, you do not need to change the current malloc to weak symbol. You only need to put your implementation (.a file) before libsgx_tstdc.a in your own enclave make file.

exFalso:
I don't want to replace the functionality, I'd like to be able to call the original function and meanwhile detect oom conditions

ZongminGu:
I don't mean replace the function too. I mean write a wrapper to add those detection code before the original malloc/free.

exFalso:
I don't have control over that code, I cannot replace the already existing malloc calls to my wrapper.

Smattr:
A less invasive way to achieve this might be to use build flags to have the [linker do this interposition for you](http://sourceware.org/binutils/docs/ld/Options.html#index-_002d_002dwrap_003dsymbol).

andyzyb:
@Smattr I was not aware of this compiler flag, but it looks a really good option.

exFalso:
@Smattr thank you, that's a much better way!

PR #88, opened 2017-03-20
ECDSA API cleanups

kaimast:
Imho this makes the ECDSA API a little cleaner.
Sorry, if there was a reason to use uint8_t instead of sgx_generic_ecresult_t.

Also, I'm not really sure why verify has two return values and if they're both needed.


andyzyb:
Any API definition change needs to consider backward compatibility (source code and binary level).  Unless there is a strong reason (wrong definition, incorrect use of types, etc.), it typically should not be changed.

kaimast:
Okay I understand. The API is just very unclear imo.

I wasn't sure if p_result was supposed to point to a buffer I allocate. I was also confused why verify takes in a non-const signature, as it shouldn't modify that signature.

kaimast:
Gonna close this.

PR #91, opened 2017-03-31
Add support for running the AESM and SGX applications in Docker

sean-jc:
This PR contains two functional modifications to the AESM to enable running SGX applications in Docker, along with scripts and files for running the AESM and SampleEnclave in separate Docker containers.

 - Bind the AESM socket to /var/run/aesmd/aesm.socket instead of an abstract socket.  Exposing sockets to Docker containers is typically done via bind mounts, which does not work for abstract sockets since they don't exist in the file system.
  - Add a --no-daemon option to the AESM service.  Application container runtimes, e.g. Docker and RKT, do not provide a program manager, e.g. systemd, inside the container, and so attempting to daemonize the AESM will fail.   The AESM will skip the call to daemon() if --no-daemon is specified.

**Note:** Users may need to remove or modify the proxy settings in the Dockerfiles, they are configured to work inside Intel's internal network.

**Note:** The Dockerfiles will not work as-is because they checkout the master branch from https://github.com/01org/linux-sgx; the PSW/SDK compiled inside the containers will not have the aforementioned changes until this PR is merged to master.  A branch that is runnable as-is can be found at https://github.com/sean-jc/linux-sgx/tree/c/docker.  The only difference is that the Dockerfiles are configured to clone and checkout https://github.com/sean-jc/linux-sgx/tree/c/docker instead of https://github.com/01org/linux-sgx.

### DIFF
--- a/psw/ae/aesm_service/source/aesm/application/main.cpp
+++ b/psw/ae/aesm_service/source/aesm/application/main.cpp
@@ -73,8 +73,16 @@ void signal_handler(int sig)
     }
 }
 
-int main() {
-    if(daemon(0, 0) < 0)
+int main(int argc, char *argv[]) {
+    // The only command line option that is supported is --no-daemon.
+    bool noDaemon = argc == 2 && (strcmp(argv[1], "--no-daemon") == 0);
+    if ((argc > 2) || (argc == 2 && !noDaemon)) {
+        AESM_LOG_INIT();
+        AESM_LOG_FATAL("Invalid command line.");
+        AESM_LOG_FINI();
+        exit(1);
+    }
+    if(!noDaemon && daemon(0, 0) < 0)
     {
         AESM_LOG_INIT();
         AESM_LOG_FATAL("Fail to set daemon.");
--- a/linux/installer/common/psw/install.sh
+++ b/linux/installer/common/psw/install.sh
@@ -54,6 +54,15 @@ chmod  0644 /etc/aesmd.conf
 chown -R aesmd /var/opt/aesmd
 chmod 0750 /var/opt/aesmd
 
+# By default the AESM's communication socket will be created in
+# /var/run/aesmd.  Putting the socket in the aesmd sub-directory
+# as opposed to directly in /var/run allows the user to create a
+# mount a volume at /var/run/aesmd and thus expose the socket to
+# a different filesystem or namespace, e.g. a Docker container.
+mkdir -p /var/run/aesmd
+chown -R aesmd /var/run/aesmd
+chmod 0755 /var/run/aesmd
+
 if [ -d /run/systemd/system ]; then
     AESMD_NAME=aesmd.service
     AESMD_TEMP=$AESM_PATH/$AESMD_NAME
@@ -112,8 +121,9 @@ $DISABLE_AESMD
 rm -f $AESMD_DEST
 rm -f /etc/aesmd.conf
 
-# Removing AESM internal folder
+# Removing AESM internal folders
 rm -fr /var/opt/aesmd
+rm -fr /var/run/aesmd
 
 # Removing runtime libraries
 rm -f /usr/lib/libsgx_uae_service.so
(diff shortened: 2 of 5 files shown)


docker/aesm/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:14.04
```
haitaohuang: it needs to work with 16.04

docker/aesm/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+ENV http_proxy http://proxy-us.intel.com:911
+ENV https_proxy http://proxy-us.intel.com:912
+ENV no_proxy intel.com,*.intel.com,127.0.0.1,localhost
+
```
haitaohuang: need remove proxy settings

docker/aesm/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get update && \
+    apt-get install -y git build-essential ocaml automake autoconf libtool libcurl4-openssl-dev libprotobuf-dev libprotobuf-c0-dev protobuf-compiler curl make g++ unzip wget libssl-dev python
+
+RUN groupadd aesmd && useradd -m -s /bin/bash -g aesmd aesmd
+
+USER aesmd
+RUN cd ~ && git clone https://github.com/01org/linux-sgx.git
+RUN cd ~/linux-sgx && \
+    ./download_prebuilt.sh && \
+    USE_OPT_LIBS=0 make psw_install_pkg
```
haitaohuang: why  USE_OPT_LIBS=0 ?

docker/sample/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM ubuntu:14.04
+
+ENV http_proxy http://proxy-us.intel.com:911
+ENV https_proxy http://proxy-us.intel.com:912
+ENV no_proxy intel.com,*.intel.com,127.0.0.1,localhost
```
haitaohuang: remove proxy and make ubuntu16.04 as base

docker/sample/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    cd /opt/intel && \
+    /home/aesmd/linux-sgx/linux/installer/bin/sgx_linux_x64_psw_1.*.bin
+
+# Install the PSW first and as a separate layer so that all layers
+# up to and including the PSW can be shared between the AESM image
+# and the SampleEnclave image.  For this same reason we use aesmd
+# as the user even though aesmd is not the most appropriate name.
+USER aesmd
+RUN cd ~/linux-sgx && \
+    USE_OPT_LIBS=0 make sdk_install_pkg
```
haitaohuang: remove  USE_OPT_LIBS=0?

llly:
It seams that Dockerfile will git clone linux-sgx and build it. So a separate linux-sgx-docker project is better than  checking Docker files in to linux-sgx project.
And you can provide a patch file in Docker project to patch linux-sgx and enhance compatibility with Docker.

sean-jc:
@llly @haitaohuang  - I dropped the last commit (Dockerfiles and run scripts) from this PR as I don't want to hold up the functional changes while we hash out the Docker configuration details.

psw/ae/aesm_service/source/aesm/application/main.cpp
```diff
@@ -73,8 +73,12 @@ void signal_handler(int sig)
     }
 }
 
-int main() {
-    if(daemon(0, 0) < 0)
+int main(int argc, char *argv[]) {
+    // The only command line option that is supported is --no-daemon,
+    // so for now we can hardcode a check against argv[1], which will
+    // be NULL if no argument is provided.
+    bool noDaemon = (strcmp(argv[1], "--no-daemon") == 0);
```
llly: must check argc before using argv. And use strncmp() instead of strcmp()

sean-jc: Using `strncmp()` is functionally incorrect and unnecessary.  Excepting a problem in the compiler or libc, in which case we have bigger issues to worry about, `--no-daemon` and `argv[1]` are guaranteed to be null terminated.  Using `strncmp()` will result in false positives, e.g. `strncmp()` will return 0 for invocations like `aesm_service --no-daemon=false` and `aesm_service --no-daemonization`.

llly: OK for strcmp.  Still must check argc before using argv.  aesm_service without any arg will coredump since arg[1] is NULL.

PR #107, opened 2017-05-22
Implements most of libsgx_capable for Linux (previously only available in the Windows SDK)

jmechalas:
Implements most of libsgx_capable for Linux (previously only available in the Windows version of the Intel SGX SDK).

The Linux implementation of this library adds the following functions:

```
sgx_status_t sgx_is_capable (int *sgx_capable);
sgx_status_t sgx_cap_get_status (sgx_device_status_t *sgx_device_status);
sgx_status_t sgx_cap_enable_device (sgx_device_status_t *sgx_device_status);
```

Adds the header file sgx_capable.h from the Windows version of the Intel SGX SDK, and also adds the following error code (also from the Windows version of the Intel SGX SDK) to sgx_error.h:

`SGX_ERROR_NO_PRIVILEGE`

Generates both a static and dynamic build of libsgx_capable. This is incorporated into the SDK package, not the PSW package, since software installers need to check for Intel SGX capability on systems where the PSW may not already be installed. This means that installers will either have to statically link, or distribute the .so with their installation package.

Signed-off-by: John P Mechalas <john.p.mechalas@intel.com>


haitaohuang:
You don't need copy the developer certificate into commit message. You certify for the compliance by signing-off.

jmechalas:
Edited the commit message.

common/inc/sgx_capable.h
```diff
@@ -0,0 +1,110 @@
...
+*           in version_string, version_string_len returns the number of characters
+*           required to hold the version string.
+*
+* @return If the function succeeds, return SGX_SUCCESS and the length of the string in
+*         version_string_len (not including NULL). SGX_ERROR_OUT_OF_MEMORY is returned if
+*         the buffer is too small or NULL, and the length required is in version_string_len.
+*         SGX_ERROR_SERVICE_UNAVAILABLE indicates the PSW is not installed. Any other value
+*         indicates an error.
+*/
+#if defined(_MSC_VER)
```
jmechalas: I assume you are referring to the sgx_cap_get_psw_version_* definitions.

I am hesitant to remove this. It's currently wrapped in an #if defined(_MSC_VER), allowing the same header file to work for both the Linux and Windows sources. If I take this out, then there has to be two separate instances of sgx_capable.h, one for Linux and one for Windows, and they'll need to be kept in sync. That sounds like more work long term for the maintainers.

sdk/libcapable/linux/sgx_capable.cpp
```diff
@@ -0,0 +1,331 @@
+#include <sys/stat.h>
```
jmechalas: Added.

sdk/libcapable/linux/Makefile
```diff
@@ -0,0 +1,76 @@
...
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+#
+
+include ../../../buildenv.mk
+
+ifndef DEBUG
+CXXFLAGS += -DDISABLE_TRACE
```
jmechalas: Removed.

common/inc/sgx_capable.h
```diff
@@ -0,0 +1,110 @@
+/**
```
andyzyb: need to change to BSD license for the header

sdk/libcapable/linux/sgx_capable.cpp
```diff
@@ -0,0 +1,355 @@
+/**
```
andyzyb: Need to change to BSD header

common/inc/sgx_capable.h
jmechalas: Ooops. Pulled the wrong header. This has been fixed.

sdk/libcapable/linux/sgx_capable.cpp
jmechalas: Ooops. Pulled the wrong header. This has been fixed.

PR #112, opened 2017-06-04
Removing sudo

mitar:
The whole script has to be run as root anyway. But sudo is not necessary installed.


llly:
Can you Squash to one commit and add a signed-off-by in the commit comment?

mitar:
Done.

Issue #100, opened 2017-04-28
sgx sdk on archlinux


n0way0ut:
there is a way to have it on distros different than ubuntu? and in particular archlinux? I tried to install it there, but there are a lot of issues..

rodolfoams:
Currently supported distros are Ubuntu, RHEL Server and CentOS.

n0way0ut:
yes, and it is quite impossible without issues to use it on different distros. Is there a plan to support other distros, or make it more generic?

andyzyb:
As there are so many Linux distros, the focus will be on several of them.  You can try to build the whole SGX software on the distro you are using and see what the issue is.  You can report the issue or submit a PR here.  If it is an easy change, it may be included.

deathholes:
I'm able to install and run sgx sdk on archlinux. Some of the issues that I recall and how I handled them are:
1) some field missing in a kernel data structure ( address field or something): I check the kernel code and found that that specific field was renamed to something else. So changed it to the new one.
2) Some error in aesm_service module (some BIO structure not found): Arch linux uses the latest version of openssl which doesn't have those files in the open(public). So I installed the package "openssl-1.0" from the extra repository and made a temporary link in the /usr/include so that it uses the header files of "openssl-1.0" instead of "openssl-1.1". 
3) installing the bin file for psw throwed an error of "arch command not found". So I changed the command from "arch" to "uname -m"
4) there were some error due to python. Arch linux links 'python' to 'python3' unlike others where it is 'python2'. Changed the code to use python2 explicitly instead of python.

Can't remember anything extra. If you still not able to build it, you can message me.

promach:
@deathholes I am still having [openssl](https://bbs.archlinux.org/viewtopic.php?pid=1885410#p1885410) issue.  Can we use the latest openssl with aesm_service module inside intel sgx ?

promach:
I managed to install both SGX and PSW using these [two PKGBUILDs](https://bbs.archlinux.org/viewtopic.php?pid=1886222#p1886222)

However, I could not [source the environment variables](https://github.com/intel/linux-sgx/blob/master/README.md#install-the-intelr-sgx-sdk-1) required for sgx code compilation.

Why is that so ?

promach:
I tried `[phung@archlinux ~]$ sudo systemctl restart aesmd
 [sudo] password for phung: 
 Failed to restart aesmd.service: Unit aesmd.service not found.
` 

[...]

PR #125, opened 2017-07-15
Fix build for more generic unix'es

Eximius:
I fixed the build here for archlinux, but it applies for other more generic unix'es as well.

I was having gcc error out due to -Werror, so some minor code changes happened.

Solves #100

### DIFF
--- a/common/inc/stdc++/linux/exception
+++ b/common/inc/stdc++/linux/exception
@@ -33,11 +33,11 @@ namespace std
 	class exception
 	{
 	public:
-		exception() throw();
-		exception(const exception&) throw();
-		exception& operator=(const exception&) throw();
+		exception();
+		exception(const exception&);
+		exception& operator=(const exception&);
 		virtual ~exception();
-		virtual const char* what() const throw();
+		virtual const char* what() const;
 	};
 
 	/**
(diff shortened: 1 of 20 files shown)


andyzyb:
Submitting a PR with multiple changes in different components is not recommended. It would be better to be split, so each change can be reviewed and decided independently.

Moschn:
Does not compile on arch. You forgot to remove the `throw()` statements in the cpp files `sdk/cpprt/memory_manage/delete2.cpp` etc.

Eximius:
Oops, fixed that.

common/inc/stdc++/new
```diff
@@ -48,22 +48,22 @@ namespace std
...
-void* SGXAPI operator new (size_t, const std::nothrow_t&) throw();
-void* SGXAPI operator new (size_t, void*) throw();
-void* SGXAPI operator new[] (size_t, const std::nothrow_t&) throw();
-void* SGXAPI operator new[] (size_t, void*) throw();
+void* SGXAPI operator new (size_t, const std::nothrow_t&);
+void* SGXAPI operator new (size_t, void*);
+void* SGXAPI operator new[] (size_t, const std::nothrow_t&);
+void* SGXAPI operator new[] (size_t, void*);
 
 void SGXAPI operator delete (void*) throw ();
```
llly: need to change it, too.

linux/installer/common/gen_source/gen_source.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
lzha101: It is not a good solution to hard code the python version.   It may be better to change the style to make it compatible.

llly: Please discard the python changes in this PR. We can change it later or in another PR.

Eximius: Well, the code in those scripts is python2, so it seems like a good idea to always make it use python2. I'll discard these.

common/inc/stdc++/new
Eximius: fixed.

common/inc/stdc++/linux/exception
```diff
@@ -33,11 +33,11 @@ namespace std
 	class exception
 	{
 	public:
-		exception() throw();
```
llly: I don't get any failures without  the changes in exception and typeinfo. Are the changes  necessary?  If necessary, exception.cc also need to be changed accordingly.

Eximius: I don't get failures either, so not necessary. But they do have exception specifications, and they are deprecated.

llly: stdexcept.cc and exception.cc in sdk/cpprt/linux should be changed according to the header changes. Thanks.

rrebane: AFAIK C++11 deprecated throw in favor of the noexcept specifier. Shouldn't "throw()" be replaced with "noexcept" instead of just removing "throw()"?

Smattr:
FYI this merged PR changed external code (TinyXML2) in a way that will cause a trivial merge conflict if/when syncing with upstream. Not a big deal, but just spotted the overlap with changes in #148 and thought it might have been overlooked.

llly:
pulled most commits in #154

PR #132, opened 2017-07-27
fix incorrect openssl API usage, which broke with >=openssl-1.1.0

geogriff:


llly:
#125 has similar changes under review.

PR #135, opened 2017-08-07
Move pms pointer validation before its dereference

deeglaze:
The "pms" pointer is validated after its use in setting up local_vars definitions.

This patch moves that validation up, and also fixes the "sgx_status_t" typo in the generated comment.


llly:
reasonable change. Since edger8r change will affect all enclaves, will be careful about when to merge.

llly:
The later 2 comments seems belong to another PR that I should merge. You can create dedicated branch for each PR.

deeglaze:
Closing this PR from the master branch to prefer the feature branch in a different PR.

PR #141, opened 2017-08-15
Check for crash in do_ocall to prevent any data leakage.

deeglaze:
The "pms" pointer is validated after its use in setting up local_vars definitions.

This patch moves that validation up, and also fixes the "sgx_status_t" typo in the generated comment.


deeglaze:
Bad copy/paste :(

PR #143, opened 2017-08-15
 Expose an internal error on enclave_create ioctl interrupt.

deeglaze:
This is an issue and PR rolled into one.

An enclave project I'm working on is getting to a semi-complicated state where the enclave .so file is about 15MiB. After a recent patch that added use of EGETKEY (red herring?), we started getting a 60% failure rate from sgx_create_enclave instead of 0%. I traced that to the ioctl call in create_enclave, where the call is failing with errno == EINTR. There is no error checking in the psw, so this patch adds a check-and-retry loop to try up to 5 times before failing with SGX_ERROR_DEVICE_BUSY. The error_driver2utrs function did not understand the -1 coming from the failing ioctl, so sgx_create_enclave was returning a less helpful SGX_ERROR_UNEXPECTED.

I don't have a small repro, and I think this is more of a band-aid solution than a real one. I thought maybe 15MiB is too big for ECREATE to validate in its time window, so I made a random enclave generator that links in /dev/urandom bytes converted through objcopy, and fiddled the enclave size up to many times our 15MiB. The sgx_enclave_create call never failed, so size can't be the only factor here.

I looked in the driver code and saw that the ioctl dispatch on enclave_create does not return EINTR, so it has to be the leaf instruction itself getting interrupted. Any clue as to why we would witness such a high failure rate?

### DIFF
--- a/common/inc/sgx_error.h
+++ b/common/inc/sgx_error.h
@@ -103,6 +103,8 @@ typedef enum _status_t
     SGX_ERROR_FILE_FLUSH_FAILED             = SGX_MK_ERROR(0x7008),	/* fflush operation (to disk) failed (only used when no EXXX is returned) */
     SGX_ERROR_FILE_CLOSE_FAILED             = SGX_MK_ERROR(0x7009),	/* fclose operation (to disk) failed (only used when no EXXX is returned) */
 
+    SGX_INTERNAL_ERROR_ENCLAVE_CREATE_INTERRUPTED = SGX_MK_ERROR(0xF001), /* The ioctl for enclave_create unexpectedly failed with EINTR. */
+
 } sgx_status_t;
 
 #endif
--- a/psw/urts/linux/enclave_creator_hw.cpp
+++ b/psw/urts/linux/enclave_creator_hw.cpp
@@ -144,6 +144,9 @@ int EnclaveCreatorHW::create_enclave(secs_t *secs, sgx_enclave_id_t *enclave_id,
     int ret = ioctl(m_hdevice, SGX_IOC_ENCLAVE_CREATE, &param);
     if(ret) 
     {
+        if(ret == -1 && errno == EINTR) {
+            return SGX_INTERNAL_ERROR_ENCLAVE_CREATE_INTERRUPTED;  // Allow the user to retry.
+        }
         SE_TRACE(SE_TRACE_WARNING, "\nSGX_IOC_ENCLAVE_CREATE failed: errno = %d\n", errno);
         return error_driver2urts(ret);
     }


lzha101:
Driver only returns EINTR from EINIT instruction.  Other ioctls are expected to restart the operations instead of returning EINTR. It is weird that you get the errno for ECREATE.  Is the failure only happens when loading your specific enclave? If yes, could you share us your enclave?  Or could you try to reproduce the failure with some modifications in SampleEnclave?

deeglaze:
Attempts at a sample repro failed. I'll have to talk to a director about sharing the binary (sorry, no source) with someone at Intel that'll sign an NDA :| do you have a contact at Intel with whom I can start this process?

lzha101:
@deeglaze, we need to double confirm with you that you get the EINTR errno during ECREATE ioctl, instead of EINIT ioctl, right? And as we have never seen this errno before, I'm not sure if this is related to your specific platform.  Is it possible that you try your enclave on some other platform of SGX supported?

deeglaze:
Confirmed that it is the ecreate ioctl. When I applied this patch to uRTS without changing the other ioctls and added logging on retry, I was able to always load my enclave with max 1 retry.

MWShan:
EINTR can be returned from SGX_IOC_ENCLAVE_INIT if 
- the EINIT instruction returns an error of SGX_UNMASKED_EVENT (this occurs when the processor gets an interrupt while processing the instruction) for several retries of the EINIT instruction
- the process/thread receives a signal

The uRTS should handle the EINTR error code for SGX_IOC_ENCLAVE_INIT, but if the error is coming from SGX_IOC_ENCLAVE_CREATE, then we need to figure out how.

Are you profiling the process while loading the enclave?  This can potentially induce interrupts which cause EINIT to return with SGX_UNMASKED_EVENT.

deeglaze:
The situation is `ioctl(m_hdevice, SGX_IOC_ENCLAVE_CREATE, &param)` returns -1 and sets `errno` to `EINTR`. It's possible that our testing harness is adding some interrupts for testing timeouts, though the timeout is for 10 minutes, not the milliseconds to failure I'm seeing. I don't know about ASAN. My attempt to delta-debug company IP out of the repro hit a different error in uRTS with EADD failing with `EINVAL` at about a 1% rate.
I tried ignoring signals around the ioctl but still see the `EINTR`.
```c
sigset_t oldset, emptyset;
sigemptyset(&emptyset);
sigprocmask(SIG_SETMASK, &emptyset, &oldset);
// <ioctl>
sigprocmask(SIG_SETMASK, &oldset, nullptr);
```

sergeay:
Would it be possible for you to do the following check - we would like to understand if the issue has to do with a call to a api: mmu_notifier_register():
1. Extract the driver code:
git clone https://github.com/01org/linux-sgx-driver.git
2. Edit sgx_ioctl.c and add the following in the error path of the call to above function (line 588):
pr_err("intel_sgx: %s failure ret=%d\n",  __func__, ret);
3. build the driver:
   make
3. stop sdk service:
   sudo service aesmd stop
4. stop driver
   sudo rmmod isgx
5. start just build driver
  sudo insmod ./isgx.ko
6. start service:
  sudo service aesmd start
7. run your scenario until the failure and send to us the sgx driver log:
  dmesg | grep intel_sgx

Thanks,

deeglaze:
What do you want for `func`? It isn't a variable, so "mmu_notifier_register"? Note: ret is long, so %d should be %ld.

haitaohuang:
mmu_notifier_register() calls mm_take_all_locks (http://elixir.free-electrons.com/linux/v4.13-rc7/source/mm/mmap.c#L3393), which can fail if it's interrupted by signals.
Is your application interrupted a lot?

deeglaze:
I'm not sure how to answer that question. Our test harness doesn't do any signalling, our driver doesn't do any signalling, and the application is single-threaded. I don't know about signals coming from the environment, but even with sigprocmask setting the block set to empty, the ioctl is interrupted by something.

sergeay:
It is not func but __func__ ; sorry
Does not matter you can remove it and remove the "%s". the important thing is to check if this call is the one that fails and the error code as well.
%d or %ld does not matter either in this case.

sergeay:
Oh oh - the editor is correcting the typing... "__func__"

sergeay:
Ok, seems you indeed hit mmu_notifier() failure.
This is unusual, what can you tell us about your environment? OS, kernel version ...
It seems your application is receiving some signals at probably a high rate.

You seem to say that this problem appeared after patching your enclave with egetkey calls; 
Again, is it possible for you to give some details without releasing confidential information?

jsakkine-intel:
How is it verified that the source for this problem is the mmu_notifier_register() call? With an ftrace filter or?

deeglaze:
@jsakkine-intel It's with the pr_err that @sergeay suggested I add at line 588.

deeglaze:
I would like to suggest that since mmu_notifier_register() is in fact interruptible, that we acknowledge that the ecreate ioctl should check for EINTR and either retry or inform the API user that they need to call sgx_create_enclave again to try. Probably it's better to retry locally, but I leave that up to you.

sergeay:
What you suggest is indeed an alternative.
We would like first to understand the root cause and to not possibly hide a bug since this behavior has not been seen until now.
In the mean time and for you to proceed I would suggest to implement temporarily your idea in the driver and replace the call to mmu_notifier_register by the following:

count = 0;
while( EINTR == (ret = mmu_notifier_register(&encl->mmu_notifier, encl->mm))) {
	count++;
	if(count >= 10)
		break;
}
if (ret) {
	encl->mmu_notifier.ops = NULL;
	goto out;
}

10 attempts when 60% Is the failure rate should be enough.

jsakkine-intel:
You must use just use `-ERESTARTSYS` instead of inventing your own homebaked ad hoc code, seriously. 

Even if you completely ignore `-EINTR` you are not really sorting out why the thread gets signaled in the first place. The cause for that is still unknown. The effect is the signal.

haitaohuang:
To follow up on comments made by @jsakkine-intel, please try this patch in driver: https://github.com/haitaohuang/linux-sgx-driver/commit/d27253869264f338e182ae916bf2a863a10d0049

Application should set SA_RESTART flag in sigactions.  (URTS has already set SA_RESTART flag in signal handler registration)

deeglaze:
I haven't yet tracked down why, but that patch along with changing the `-EINVAL` to `ret` at https://github.com/01org/linux-sgx-driver/blob/master/sgx_ioctl.c#L700 (another problem with signal_pending) makes the `ioctl` calls stop failing, but the `sgx_create_enclave` call still fails with `SGX_ERROR_UNEXPECTED`.

chrisr3:
Heh, so much effort just to write `__func__` ... ;-)

deeglaze:
I've brought this PR back up to date with HEAD.

haitaohuang:
Do you still see this issue with latest driver and uAE?
We did several patches related to signal handling in both uAE and driver and no long was able to reproduce the issue with the tests. 
see discussions in https://github.com/intel/linux-sgx/issues/158

deeglaze:
I've updated our driver to v2.1 and ran from SDK v1.9 with PR168 downstreamed and still see intermittent creation failure. Are there other explicit patches you can point me to that might help?

sergeay:
The driver is now returning -ERESTARTSYS such that the kernel should do the required retries and you still get those errors? We need to get to the root cause of this issue.
I understand that you can not share your so file, but can you maybe remove the part that is critical for you and with some chance, the problem will reproduce

haitaohuang:
@deeglaze Yes those are the patches. patches more relevant to this particular issue (EINTR returned from ECREATE IOC, not from other IOC or UAE APIs) are in the driver. As you still see EINTR returned in latest driver which is supposed to auto restart the IOC,  I agree with @sergeay there may be another driver/kernel issue. It'd be desirable if we can track down the root cause.

couple of things I would suggest like to check:

1) All application  sigactions has SA_RESTART flag set when registering signal handling
2) I have this test to emulate random signals sent to an app that loops on sgx_create_enclave: https://github.com/haitaohuang/linux-sgx/commit/7a195f17a9a0951ecf1462dc264da893d4bf0b19?
Can you try that in your environment?

If by any luck it actually reproduces this issue, then we can compare our environment settings to see what's going on.
If it does not reproduce the failure, it also confirms the issue is different than what we have fixed

sergeay:
I just tried the above w/o any success, i.e. no repro and 100% create_enclave() success .
I am wondering from where this EINTR is coming from now. 
Maybe we do not understand completely the ERESTARTSYS mechanism.
Do you know the size of your enclaves? I am wondering if eviction  is working during your process.

sergeay:
@deeglaze, I would like to be sure; are you positive that with latest driver code you are still getting -EINTR return code from ecreate ioctl?
If this is the case, I would like to know which system call does return that value, that would be very easy if I had access to your environment but if you could add some kernel prints there, that would definitely help to find the root cause.

deeglaze:
The size appears to not have anything to do with it. Some months ago I wrote a test that made an enclave that links in however much random garbage I ask for and prints a random character from it. No matter the size (within a factor of 4 of the problem enclave), I was unable to repro.

PR #146, opened 2017-08-22
Fix potential false alarm of stack smashing attack

tatetian:
This PR makes sure all TRTS functions during which the stack guard (or canary value) may be initialized are compiled without `-fstack-protector-strong`. Currently, functions (in `trts_nsp.cpp`) that directly deal with stack guard initialization are correctly compiled without `-fstack-protector-strong`; yet, the callers of these functions (`do_ecall()` in `trts_ecall.cpp` and `enter_enclave()` in `trts.cpp`) are not. This makes the code of `enter_enclave()` and `do_ecall()` very fragile (see the example below) since any code changes that cause the compiler to add stack protection logic for these two functions would break them. This PR simply moves the code of `do_ecall()` and `enter_enclave()` to `trts_nsp.cpp`. As the file name of `trts_nsp.cpp` suggests, it is where No-Stack-Protection code are supposed to be anyway.

Here is an example of how seemingly innocent code changes could break `do_ecall()`, causing Illegal Instruction error upon the first ECall to any enclave compiled:

    diff --git a/sdk/trts/trts.cpp b/sdk/trts/trts.cpp
    index 9bbdb5e..a909117 100644
    --- a/sdk/trts/trts.cpp
    +++ b/sdk/trts/trts.cpp
    @@ -321,3 +321,7 @@ extern "C" int enter_enclave(int index, void *ms, void *tcs, int cssa)
         }
         return error;
     }
    +
    +extern "C" int reproduce_bug(int* a) {
    +    return *a;
    +}
    diff --git a/sdk/trts/trts_ecall.cpp b/sdk/trts/trts_ecall.cpp
    index 1de3cfc..310b881 100644
    --- a/sdk/trts/trts_ecall.cpp
    +++ b/sdk/trts/trts_ecall.cpp
    @@ -124,6 +124,9 @@ static sgx_status_t trts_ecall(uint32_t ordinal, void *ms)
         sgx_status_t status = get_func_addr(ordinal, &addr);
         if(status == SGX_SUCCESS)
         {
    +        int a = 1;
    +        reproduce_bug(&a);
    +
             ecall_func_t func = (ecall_func_t)addr;
             status = func(ms);
         }
    diff --git a/sdk/trts/trts_internal.h b/sdk/trts/trts_internal.h
    index 28c0f48..bcbb54e 100644
    --- a/sdk/trts/trts_internal.h
    +++ b/sdk/trts/trts_internal.h
    @@ -72,6 +72,9 @@ sgx_status_t do_init_enclave(void *ms);
     sgx_status_t do_ecall(int index, void *ms, void *tcs);
     sgx_status_t do_oret(void *ms);
     sgx_status_t trts_handle_exception(void *tcs);
    +int reproduce_bug(int* a);
     #ifdef __cplusplus
     }
     #endif

Use this patch with `git apply` on any commits since 1.9 release, compile the SDK, and try any project under `SampleCode`. The sample enclaves will always report Illegal Instruction error.

### DIFF
--- a/sdk/trts/trts.cpp
+++ b/sdk/trts/trts.cpp
@@ -38,8 +38,6 @@
 #include "thread_data.h"
 #include "global_data.h"
 
-#include "internal/rts.h"
-
 #ifdef SE_SIM
 #include "t_instructions.h"    /* for `g_global_data_sim' */
 #include "sgx_spinlock.h"
@@ -285,39 +283,4 @@ sgx_status_t sgx_read_rand(unsigned char *rand, size_t length_in_bytes)
     }
     memset_s(&rand_num, sizeof(rand_num), 0, sizeof(rand_num));
     return SGX_SUCCESS;
-}
-
-#include "trts_internal.h"
-extern "C" int enter_enclave(int index, void *ms, void *tcs, int cssa)
-{
-    if(get_enclave_state() == ENCLAVE_CRASHED)
-    {
-        return SGX_ERROR_ENCLAVE_CRASHED;
-    }
-
-    sgx_status_t error = SGX_ERROR_UNEXPECTED;
-    if(cssa == 0)
-    {
-        if(index >= 0)
-        {
-            error = do_ecall(index, ms, tcs);
-        }
-        else if(index == ECMD_INIT_ENCLAVE)
-        {
-            error = do_init_enclave(ms);
-        }
-        else if(index == ECMD_ORET)
-        {
-            error = do_oret(ms);
-        }
-    }
-    else if((cssa == 1) && (index == ECMD_EXCEPT))
-    {
-        error = trts_handle_exception(tcs);
-    }
-    if(error == SGX_ERROR_UNEXPECTED)
-    {
-        set_enclave_state(ENCLAVE_CRASHED);
-    }
-    return error;
 }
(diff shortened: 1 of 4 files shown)


ZongminGu:
The PR kills the protection of a critical function: do_ecall. The benefit is to support modification of this api, and the modification is not necessary. I think it is better to accept this PR when we really need add local variables in do_ecall.

tatetian:
Thanks @ZongminGu for the prompt reply.

This PR does not kill the stack protection for `do_ecall()` in the sense that 1) the current implementation of  `do_ecall()` (or `enter_enclave()`) does not meet [the condition of generating stack protection code by compiler](https://docs.google.com/document/d/1xXBH6rRZue4f296vGt9YQcuLVQHeE516stHwt8M9xyU/edit) ; 2) all future implementation of `do_ecall` (or `enter_enclave()`) *cannot* use stack protection as long as the function calls `do_init_thread()`, which calls `init_stack_guard()` in turn.

I encountered this issue when I tried to do some modification to `do_ecall()`. Other contributors may encounter this issue, too. It's very *surprising* to see simple, benign modification to `do_ecall()` or `enter_enclave()` can cause fatal error at runtime. This PR can save the hours of debugging time for any developers who are not familiar with how stack protection works inside enclaves. In addition, by putting all functions that cannot use stack protection mechanism to a central place (`trts_nsp.cpp`), this PR makes the code more clear and consistent with itself.

There is nothing to lose with this PR. It only improves the quality of code.

ZongminGu:
In the next release we have a different solution to address this issue instead of put functions into the _nsp.cpp.

lzha101:
Hi @tatetian,  the conclusion of nothing to lose with this PR is gotten from the comparison with current implementation, because current implementation for do_ecall()/enter_enclave() do not generate stack protection.  Just as Zongmin said, we will deliver a better implementation to fix this for next release. Please stay tuned.

tatetian:
Hi @ZongminGu, @lzha101, thanks for the clarification. As you already have plan to address this issue, I am closing this PR. Cheers.

PR #148, opened 2017-08-24
Fix build issues on g++-7

kaimast:
Throw specifiers (except nothrow) shall not be used anymore. Modern versions of g++ create an error message about that.
Also, fallthroughs in switch statements need to explicitly specified now so the compiler does not complain.

### DIFF
--- a/SampleCode/Cxx11SGXDemo/README.txt
+++ b/SampleCode/Cxx11SGXDemo/README.txt
@@ -18,7 +18,7 @@ The project demonstrates serveral C++11 features inside the Enclave:
 - function object wrapper;
 - atomic, mutexes, condition_variables;
 - new smart pointer classes: shared_ptr, unique_ptr;
-- new c++ algorithms: all_off, any_of, none_of;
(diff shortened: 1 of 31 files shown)


Smattr:
Part of this seems to change third-party code. Looks as if the external [addressed this](https://github.com/leethomason/tinyxml2/commit/3cc6f5c6f2184ed56fb13353ef5d03d38f00dac6) in a different way that's backwards compatible with non-C++11 compilers. It might be simpler in the long term to resync with upstream instead.

chrisr3:
Are you sure this is correct? GCC 7.2 only complains about `throw(std::bad_alloc)` on two `new` operators for me, and it also requires that I remove them from _both_ the prototypes _and_ the C++ function bodies.

llly:
@chrisr3 Agree. C++ function bodies should be changed.

kaimast:
Yeah you're right. The makefile in the SDK is a little buggy and didn't actually rebuild the cpp file when I changed the header.

Pushed the cpp stuff too.

PR #165, opened 2017-09-18
Ensure trts_nsp.cpp is compiled without stack protection.

chrisr3:
This file must _not_ be compiled with `-fstack-protector-strong`. Removing the compiler option is not enough because the compiler may apply `-fstack-protector-strong` anyway by default. We _need_ to specify an explicit compiler option to disable it properly.

### DIFF
--- a/sdk/trts/Makefile
+++ b/sdk/trts/Makefile
@@ -40,7 +40,7 @@ CXXFLAGS += $(ENCLAVE_CXXFLAGS) \
             -fno-exceptions \
             -fno-rtti
 
-TCXXFLAGS := $(filter-out -fstack-protector-strong, $(CXXFLAGS))
+TCXXFLAGS := $(patsubst -fstack-protector%,-fno-stack-protector,$(CXXFLAGS))
 
 OBJS1 := init_enclave.o  \
         trts.o           \


lzha101:
Thanks. Please make sure you agree to the [License.txt](https://github.com/01org/linux-sgx/blob/master/License.txt) and  add a Signed-off-by to your commit message according to [CONTRIBUTING.md](https://github.com/01org/linux-sgx/blob/master/CONTRIBUTING.md).

lzha101:
BTW, to definitely disable the stack check for this file, how about using `-fno-stack-protector`?
````
diff --git a/sdk/trts/Makefile b/sdk/trts/Makefile
index 840fac2..a9b12c1 100644
--- a/sdk/trts/Makefile
+++ b/sdk/trts/Makefile
@@ -40,7 +40,7 @@ CXXFLAGS += $(ENCLAVE_CXXFLAGS) \
             -fno-exceptions \
             -fno-rtti

-TCXXFLAGS := $(filter-out -fstack-protector-strong, $(CXXFLAGS))
+TCXXFLAGS := $(patsubst -fstack-protector%,-fno-stack-protector,$(CXXFLAGS))

 OBJS1 := init_enclave.o  \
         trts.o           \
````

chrisr3:
Changed to use `-fno-stack-protector`, which is more consistent with the file suffix `nsp` anyway.

PR #176, opened 2017-10-19
Mention environment set up to compile samples

clenimar:
In some cases, it may be necessary to source the needed environment
variables before compiling the sample codes. Mention it in the main
README and also in every sample code README.

Even if the installer tells the user to source these vars, it's an easy thing to miss
and I think it should be mentioned in the repo documentation.

Signed-off-by: Clenimar Filemon <clenimar.filemon@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,10 @@ To install the Intel(R) SGX SDK, invoke the installer, as follows:
 $ cd linux/installer/bin
 $ ./sgx_linux_x64_sdk_${version}.bin 
 ```
+NOTE: You need to set up the needed environment variables before compiling your code. To do so, run:
+```
+   $ source ${sgx-sdk-install-path}/environment
+```
 ### Test the Intel(R) SGX SDK Package with the Code Samples
 - Compile and run each code sample in Simulation mode to make sure the package works well:    
 ```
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -161,7 +162,13 @@ See the later topic, *Install Intel(R) SGX PSW*, for information on how to insta
   $ make
   $ ./app
 ```
-   Use similar commands for other code samples.   
+   Use similar commands for other code samples.
+
+   NOTE: You may need to set up the needed environment variables before compiling the code sample. To do so, run:
```
yuyuany: In the main readme, we can mention the note in the "Install the Intel(R) SGX SDK" step, because it is not only required by sampe code, but also other enclaves.

README.md
```diff
@@ -161,7 +162,13 @@ See the later topic, *Install Intel(R) SGX PSW*, for information on how to insta
...
+```
+   $ source ../../environment
```
yuyuany: User can install the sdk to other places, so we'd use ${sgx-sdk-install-path} here. Same comments for the other use of the relative path of 'environment'

clenimar:
thanks for your suggestions, @yuyuany. just re-rolled the PR.

SampleCode/Cxx11SGXDemo/README.txt
```diff
@@ -26,7 +26,9 @@ The project demonstrates serveral C++11 features inside the Enclave:
 How to Build/Execute the C++11 sample program
 ---------------------------------------------
 1. Install Intel(R) Software Guard Extensions (Intel(R) SGX) SDK for Linux* OS
-2. Build the project with the prepared Makefile:
+2. Make sure your environment is set:
+    a. source ${sgx-sdk-install-path}/environment
```
yuyuany: change 'a.' to '$'. Same comments for other files.

clenimar: done.

PR #185, opened 2017-11-29
Update script to look into specific SGXSSL tag.

akryeem-INTC:
Instead of downloading and using master branch, use specific tag that we know will always work for SGX SDK.
When SGXSSL master branch change, SGX SDK should be affected.


lzha101:
@akryeem-INTC, it looks good for me. Seems you missed the Signed-off-by in the commit message.  Will merge the change after you add that.

PR #202, opened 2018-01-25
Add `-unsafe-string` flag to ocamlbuild.

samscott89:
ocamlbuild by default uses `-safe-string`, override that flag to allow most recent version to work.

See [here](https://caml.inria.fr/pub/docs/manual-ocaml/libref/String.html) for more information.


llly:
It's not a good idea to disable any security feature. It's a workaround when using recent version Ocaml. We will update Edger8r implementation to support Ocaml 4.01 to 4.06 in next release.

samscott89:
Okay, that makes sense. Thanks.

PR #203, opened 2018-01-26
Fix OpenSSL 1.1.0 compatibility issue in sign_tool and urts_sim

lzha101:


light1021:
Hi，Thanks for your replay. I missed that.

PR #210, opened 2018-02-16
Remove #defines reserved by the C standard

deeglaze:
The libc implementation newlib defines these symbols and causes a compiler error. Per the standard (ISO 9899:1999 reserved identifiers section), any identifier beginning with two underscores or an underscore and a capital letter is reserved for libc. The definitions are not used anywhere in the source code either.


haitaohuang:
looks good to me but lacks sign-off

deeglaze:

deeglaze:
Rebased on head and signed off.

deeglaze:
Ping :)

PR #211, opened 2018-02-17
Removes the use of an optional C11 extention

deeglaze:
The use of errno_t causes an unnecessary headache with gcc toolchains.


deeglaze:
This is an incomplete PR. Closing for now.

deeglaze:
All errno_t occurrences removed.

deeglaze:
Really having a hard time with git history here...

PR #212, opened 2018-02-17
Removes errno_t uses since it's painful to use in some toolchains.

deeglaze:
The C11 standard has errno_t as an optional addition which is not as widely supported as users of this SDK would hope.
Appendix K already defines errno_t as int, so this PR does the replacement.

### DIFF
--- a/SampleCode/RemoteAttestation/service_provider/ecp.cpp
+++ b/SampleCode/RemoteAttestation/service_provider/ecp.cpp
@@ -40,7 +40,7 @@
 
 #define MAC_KEY_SIZE       16
 
-errno_t memcpy_s(
+int memcpy_s(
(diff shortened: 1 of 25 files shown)


lzha101:
Could you tell us your compiling environment?  Curious why you need to replace all the errno_t including both untrusted and trusted components, as the trusted code should be compiled without involving the standard header files.

deeglaze:
It's an older version of gcc. This patch is overzealous because errno_t is non-standard and unnecessarily obnoxious.

lzha101:
@deeglaze, one question for more details - Does the type errno_t block your build?

deeglaze:
We've modified our added runtime support to define errno_t. We no longer need this PR.

PR #215, opened 2018-02-23
Adds an API function to check if the enclave has crashed.

deeglaze:
Since thread A in an enclave can't be preempted by thread B (which calls `abort()`), this API call allows for cooperation to shut down more gracefully. An OS may not tear down a process immediately, so it's useful for threads to periodically check if they're in an aborted state and shut down.


haitaohuang:
Looks good to me. Could you add sign-off line?

deeglaze:

lzha101:
@deeglaze , you need to add your signed-off to the commit log message instead of adding it here. See [CONTRIBUTING.md](https://github.com/intel/linux-sgx/blob/master/CONTRIBUTING.md) for details. Perhaps `git commit **-s**` could help you to do this.

deeglaze:
Done.

PR #217, opened 2018-02-23
Changes continue_execution to not clobber stack that may be in use.

deeglaze:
We use the exception handler to implement cpuid. To get that to work we have to ensure more of the stack is available. `continue_execution` should not store the instruction address for resumption in a part
of the stack that is likely in use. Otherwise, it stores it at the next stack location, which at least OPENSSL_cpuid has in use at the time of CPUID execution, corrupting important information.


lzha101:
Do you have a sample to reproduce this?  The `ssa_gpr->xsp` saves the stack pointer of the function in which the exception happens. Suppose the code should not save data to the stack area that above the top of the stack.   So it should be OK for trts exception handling to setup a new stack frame for `continue_execution` and save the ip on it.  It is weird that your code needs to access the stack area above the top of the stack.

BTW, are you trying to use OpenSSL inside enclave? FYI, we have an [SGXSSL](https://github.com/intel/intel-sgx-ssl) for that use.

deeglaze:
We're using BoringSSL, Google's vetted SSL library. It weirdly uses stack beyond RSP. I'll confer with my team about what exactly is the problem and update you tomorrow.

deeglaze:
This change is to support programs that are compiled with stack red zones. I can see about making a repro test if y'all need.

deeglaze:
I pushed an update to this PR that explains the red zone situation a bit more. I think at the very least internal_handle_exception should be annotated with __attribute__((interrupt)) to play nicely with red zones when the compiler can handle it.

lzha101:
The changes may cause problem for 32 bit mode as 32 bit doesn't have red zone.  Could you help verify if this commit works for you?   https://github.com/lzha101/linux-sgx/commit/f361038eba15d2bdfe74c59f92b851b80c120d53

deeglaze:
That commit causes a crash for us. Externalizing a repro would probably be more work than just roundtripping with me at this time, unfortunately.

lzha101:
It's weird. Of cause, could you provide a test repo?

yuyuany:
A potential problem of the patch is that continue_execution is running on a static stack, whose size is only no more than 1 page. We need to make sure the extra 128 byte doesn't overflow the stack. Otherwise, the stack used by ECALLs  will be overwritten.

deeglaze:
Okay, I just double and triple checked y'all's suggested patch and it indeed works for us. Sorry for the mixup.

lzha101:
I create another PR #231 for the red zone fix. Could you help review it? If it is OK, we will merge it later.

PR #222, opened 2018-03-08
Add exception handler to edger8r.

kongoshuu:
Add a exception_handler to the edger8r. If any function in .edl file is marked
as "exception_handler", it should be set the "is_exception" field in the ecall
table. This is used for users to set their own exception, like signal
handler.

### DIFF
--- a/sdk/edger8r/linux/Ast.ml
+++ b/sdk/edger8r/linux/Ast.ml
@@ -159,6 +159,7 @@ type func_decl = {
 type trusted_func = {
   tf_fdecl   : func_decl;
   tf_is_priv : bool;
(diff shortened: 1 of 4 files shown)


ZongminGu:
What is the motivation to high light the exception function in the edl file?

kongoshuu:
This is used when a user what to register their own exception handler. They can design their own exception handling that enters an enclave by highlighting the ecall as "exception_handler". Another PR has the changed code that deals with this flag (#223 ).

ZongminGu:
The alternative solution is, when the application get the SIGCHLD, SIGUSR1, SIGRTMIN + n, etc, the application can use another free thread to handle this exception.

PR #227, opened 2018-03-12
Change sgx STATIC_STACK_SIZE

kongoshuu:
The current size is causing problems when more than one frame exists inside
enclave, since very limited static stack size is provided for that frame, which
may cause stack overflow for the user exception handling feature we added here.
This is a temporary solution, a better solution is to be added to handle stack.

Also increase the minimum enclave stack size to be 0x3000 to be greater than
this.

Signed-off-by: Chong Cai chongc@google.com


andyzyb:
Is this change still needed?

andyzyb:
close since no further response.

PR #231, opened 2018-03-16
Fix red zone issue in continue_execution()

lzha101:
Also put some macros shared in C/C++ and assembly code to one header file.

### DIFF
--- a/sdk/trts/linux/trts_pic.S
+++ b/sdk/trts/linux/trts_pic.S
@@ -552,8 +552,10 @@ DECLARE_LOCAL_FUNC continue_execution
     push    %xax                       # push xax
     mov     SE_WORDSIZE*1(%xcx), %xax
     push    %xax                       # push xcx
-    mov     SE_WORDSIZE*4(%xcx), %xax
-    sub     $(SE_WORDSIZE), %xax       # xax: xsp
+    mov     SE_WORDSIZE*4(%xcx), %xax  # xax: xsp 
+# x86_64 requires a 128-bytes red zone. We need to allocate buffer to avoid touching the red zone.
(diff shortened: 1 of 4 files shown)


sdk/trts/linux/trts_pic.S
```diff
@@ -553,7 +553,7 @@ DECLARE_LOCAL_FUNC continue_execution
     mov     SE_WORDSIZE*1(%xcx), %xax
     push    %xax                       # push xcx
     mov     SE_WORDSIZE*4(%xcx), %xax
-    sub     $(SE_WORDSIZE), %xax       # xax: xsp
+    sub     $(SE_WORDSIZE + RED_ZONE_SIZE), %xax       # xax: xsp
```
deeglaze: Change comment to address the added value?

sdk/trts/linux/trts_pic.h
```diff
@@ -39,6 +39,7 @@
 
 #include "linux/linux-regs.h"
 #include "rts_cmd.h"
+#include "trts_common_macro.h"
```
deeglaze: "common_macro" is a non-suggestive name. Do you have plans to grow it outside of stack constants? Maybe trts_stack_constants.h would be better here.

lzha101: Actually we adding this header file is to put the macros shared between assembly code and C/C++ code put into one header file to avoid defining twice.  trts_stack_constants.h may not be good because we may have other shared constants which is not for stack.

lzha101: I update the patch to refine the comments and rename the header file as 'trts_shared_constants.h'.

PR #234, opened 2018-03-26
Annotate fallthroughs for -Wimplicit-fallthrough.

deeglaze:
GCC 7 and clang have an analysis that warns against implicit fallthroughs. This PR adds an annotation to make the fallthroughs explicit. C++17 adds [[fallthrough]] as a statement annotation, but until then, clang- and gcc-specific annotations will have to do.


deeglaze:
These annotations actually work less readily than the existing /* fallthrough */ comments.

PR #273, opened 2018-06-05
remove dup mov in load_regs

daveti:


yuyuany:
The patch looks fine. But could you read CONTRIBUTING.md and add signed-off-by to the commit message, so that we can merge the PR.

daveti:
Does this work?

PR #286, opened 2018-06-27
Update checksum in the download script.

lzha101:
Update checksum for the re-uploaded prebuilt binaries.


lzha101:
Will merge this patch after the updated package synced to remote server later.

andyzyb:
Should this be merged as pre-built binaries are updated?

PR #301, opened 2018-08-10
Localize -O0 to necessary functions

deeglaze:
The load_regs assembly function uses its caller's frame pointer.
C/C++ compilers are allowed to make optimizations to their
representation of the stack, and frequently omit the frame pointer.

The functions _EEXIT in t_instructions.cpp and _SE3 in
u_instructions.cpp both call load_regs, and thus should be the only ones
that should be affected by deoptimization.

Even -O0 may omit the frame pointer (indeed we found this to be the case
when compiling with GCC 7.3.0), so this change also makes that requirement
explicit by by means of -fno-omit-frame-pointer.

An added bonus of switching the -O0 flag over to an attribute in source
code is that alternate build tools are easier to port to :)

### DIFF
--- a/sdk/simulation/uinst/Makefile
+++ b/sdk/simulation/uinst/Makefile
@@ -63,7 +63,7 @@ enclave_mngr.o: enclave_mngr.cpp
 # since the '_SE3' function has assumptions on stack layout.
 #
 u_instructions.o: u_instructions.cpp
-	$(CXX) $(CXXFLAGS) $(CPPFLAGS) -O0 -Wno-error=cpp -c $< -o $@
+	$(CXX) $(CXXFLAGS) $(CPPFLAGS) -Wno-error=cpp -c $< -o $@
 
 $(LIBSESIMU_U): u_instructions.o enclave_mngr.o $(OBJ1)
 	$(AR) rcs $@ $^
--- a/sdk/simulation/assembly/lowlib.h
+++ b/sdk/simulation/assembly/lowlib.h
@@ -40,6 +40,15 @@
 
 #include <stdint.h>
 
+#if defined(__GNUC__) && !defined(__clang__)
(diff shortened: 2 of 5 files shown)


sdk/simulation/uinst/Makefile
```diff
@@ -63,7 +63,7 @@ enclave_mngr.o: enclave_mngr.cpp
 # since the '_SE3' function has assumptions on stack layout.
 #
 u_instructions.o: u_instructions.cpp
-	$(CXX) $(CXXFLAGS) $(CPPFLAGS) -O0 -Wno-error=cpp -c $< -o $@
+	$(CXX) $(CXXFLAGS) $(CPPFLAGS) -Wno-error=cpp -c $< -o $@
```
ZongminGu: What is the reason the modification is different from sdk/simulation/tinst/Makefile?

sdk/simulation/uinst/td_mngr.h
```diff
@@ -56,6 +56,15 @@ typedef struct _tcs_sim_t
 #define GET_TCS_PTR(xbp) (tcs_t *)(*(uintptr_t *)((size_t)(xbp) + 2 * sizeof(uintptr_t)))
 #endif
 
+#if defined(__GNUC__) && !defined(__clang__)
```
lzha101: It may not be proper to add the macro to this file as it is only for td/tcs related.  How about adding them to sdk/simulation/assembly/lowlib.h?

sdk/simulation/uinst/Makefile
deeglaze: I removed -O0 from both. The makefiles don't look the same, so the implementation of removing -O0 looks different.

sdk/simulation/uinst/td_mngr.h
deeglaze: Moved.

Issue #246, opened 2018-04-10
Rationale for pseudo-random numbers in simulation mode


kaimast:
Hi,

Is there a specific reason why sgx_read_rand returns a pseudo-random number in simulation mode? 
(as document here https://software.intel.com/en-us/node/709094)

In particular it results my application to always generate the same public/private key pair in simulation mode which causes various problems.

Is there any way to force randomness in simulation mode?

Thanks,
Kai

ZongminGu:
I tested it in a simple case:

65 void test()
 66 {
 67     int a = 0;
 68
 69     for (int i=0; i<4; ++i)
 70     {
 71
 72     sgx_status_t status = sgx_read_rand((unsigned char *)&a, sizeof(a));
 73
 74     printf("random:%d\n", a);
 75
 76 }
 77 }

I run it twice. The result is:
./app
random:-1514151190
random:793048459
random:-522582890
random:168439143
Enter a character before exit ...

./app
random:308955192
random:1344177320
random:-218509240
random:1266178082
Enter a character before exit ...

The value changed every time.

kaimast:
We don't use simulation in production but for our automated tests. Hopefully in the future we can test in prerelease mode but it's just not possible with the current virtualization infrastructure. 

So it seems that the seed for randomness is derived from the local time. Is that true? I added a delay in my test suite between starting enclaves and it seems to fix the problem.

sergefdrv:
@kaimast I encountered the same effect. It looks like the random number generation is pre-seeded with current time in seconds. So in all tests, I have to add one second delay between invocations to make sure that it is unlikely to get the same random value... That makes unit-testing awfully slow :slightly_frowning_face:

sergefdrv:
I believe this line is the guilty: https://github.com/intel/linux-sgx/blob/54cae063cd0d21be5fab28c0d4b81b073b5d0914/sdk/simulation/urtssim/enclave_creator_sim.cpp#L232

It would be better to use nanoseconds for the seed value.

lzha101:
In the latest release, we have changed the random generator on simulation mode to use RDRAND if the platform supports RDRAND instruction to reduce the usage of LCG.  
https://github.com/intel/linux-sgx/blob/master/sdk/trts/trts.cpp#L258

And we suggest to use HW mode instead of simulation mode. Simulation mode doesn't support all the SGX features.

sergefdrv:
[...]

PR #304, opened 2018-08-21
seed with nanosecond timestamp in simulation mode

Naoya-Horiguchi:
This pull request is trying to solve the issue #246.

ChangeLog v1->v2:
- remove inclusion of <time.h> (per @sergefdrv)

ChangeLog [v2](https://github.com/intel/linux-sgx/compare/master...Naoya-Horiguchi:post/nanosecond_seeding.v2)->v3:
- use `clock_gettime()` instead of `high_resolution_clock`

--- quoting patch description below ---
Currently we seed random number generator with timestamp in seconds.
Because of this we have to insert 1 second delay in each testcase to
ensure the randomness, which prevents us from testing applications
effeciently.

Simulation mode is not supposed to be used in production, but
developing applications on environtment without SGX HW is still common.
So let's seed with timestamp in nanoseconds in simulation mode.

Signed-off-by: Naoya Horiguchi <n-horiguchi@ah.jp.nec.com>

### DIFF
--- a/sdk/simulation/urtssim/enclave_creator_sim.cpp
+++ b/sdk/simulation/urtssim/enclave_creator_sim.cpp
@@ -229,7 +229,9 @@ int EnclaveCreatorSim::initialize(sgx_enclave_id_t enclave_id)
     assert(global_data_sim_ptr != NULL);
 
     // Initialize the `seed' to `g_global_data_sim'.
-    global_data_sim_ptr->seed = (uint64_t)time(NULL);
+    struct timespec ts;
+    clock_gettime(CLOCK_REALTIME, &ts);
+    global_data_sim_ptr->seed = (uint64_t)ts.tv_sec * 1000000000ULL + (uint64_t)ts.tv_nsec;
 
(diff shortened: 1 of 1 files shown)


Naoya-Horiguchi:
I received notfications from @sergefdrv's comments, which seems unreachable on web UI maybe because I did force-update of the pull request branch. So let me share them here:

> Given my mostly C-only background, I would most likely have used clock_gettime.
> This C++11 way of doing things looks somewhat complicated to me  But if the
> maintainers of the project are fine then I should be too ^J

I'm fine on using clock_gettime.

> I think it's better to change double quotes to angle brackets here, since
> chrono is a standard header file.

With clock_gettime, we don't need change header part, so the patch will be simple.

Anyway I'll wait a little for maintainer's feedback.

lzha101:
If clock_gettime() could fix your problem,  using C implementation may be better as to the readability for more developers.  And it is more simple than C++ implementation.

Naoya-Horiguchi:
@lzha101, @sergefdrv thanks for comments. I've updated the branch using `clock_gettime()`, which would be the better fix.

PR #307, opened 2018-08-27
Remove an unused #include

deeglaze:
This was committed upstream at https://github.com/gperftools/gperftools/commit/977e0d45003514d088e4bb83474c985094acee47

### DIFF
--- a/sdk/gperftools/gperftools-2.5/src/base/vdso_support.cc
+++ b/sdk/gperftools/gperftools-2.5/src/base/vdso_support.cc
@@ -43,7 +43,6 @@
 #include <stddef.h>   // for ptrdiff_t
 
 #include "base/atomicops.h"  // for MemoryBarrier
-#include "base/linux_syscall_support.h"
 #include "base/logging.h"
 #include "base/dynamic_annotations.h"
 #include "base/basictypes.h"  // for COMPILE_ASSERT


sdk/gperftools/gperftools-2.5/src/base/vdso_support.cc
```diff
@@ -43,7 +43,6 @@
 #include <stddef.h>   // for ptrdiff_t
 
 #include "base/atomicops.h"  // for MemoryBarrier
-#include "base/linux_syscall_support.h"
```
andyzyb: Does this change come from upstream?

deeglaze: It does not. I mistakenly thought all third-party dependencies were in the external/ directory.

deeglaze: Oh, but I see they had the same idea https://github.com/gperftools/gperftools/commit/977e0d45003514d088e4bb83474c985094acee47#diff-2599408afdb3c2829102d83ae47df1cb

andyzyb: Is this patch accepted by upstream, if so we can merge it.

PR #308, opened 2018-09-05
Include sl_util.h to define MIN in mini_snprintf.h

deeglaze:
The MIN macro is not defined in any of the included headers.
Since it's only used in one place, replace the use with the intended
expansion instead of defining MIN.


lzha101:
This header file is added to support  Switchless Calls. And there is a MIN definition in the header file (common/inc/internal/sl_util.h) for switchless.  How about including the header file to this header?

deeglaze:
Updated.

PR #309, opened 2018-09-05
Remove #include sgx_urts.h from sgx_uswitchless

deeglaze:
The sgx_uswitchless.h header is used in both trusted and untrusted
contexts from transitive inclusions. The sgx_urts.h header is strictly
an untrusted domain header. Users that depend on the untrusted runtime
should not depend on the transitive include behavior to provide
sgx_urts.h.


andyzyb:
Looks good.  Will wait for a better timing for merging the change in the coming week.

PR #310, opened 2018-09-11
fix #177

### LINKED ISSUE #177: SGX SDK installer is x86,not x64
Hello,When I use command
```
make sdk_install_pkg
```
to build the Intel(R) SGX SDK installer,the installer should be x64,but in my computer is x86, 
the install as follows:
image

and the information of my computer as follows:
image

jat001:
The colon is fullwidth in CJK computing. Remove it for compatibility.

### DIFF
--- a/linux/installer/bin/build-installpkg.sh
+++ b/linux/installer/bin/build-installpkg.sh
@@ -57,7 +57,7 @@ BUILD_DIR=${ROOT_DIR}/build/linux
 # Get the architecture of the build from generated binary
 get_arch()
 {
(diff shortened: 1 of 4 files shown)


llly:
Your patch works. We'll merge it a few days later.

lzha101:
In the latest release, the file `linux/installer/common/libsgx-enclave-common/createTarball.sh` is added and also needs to be updated with the same change.  Could you please update your patch to include the change? Then we can merge it. Thank you.

jat001:
@lzha101 updated

PR #316, opened 2018-09-20
Allow enclaves to be created from buffers

birdperson-lives:
To that end, also allow the m_enclave_info.lpFileName field of a loaded CEnclave to be empty, since an enclave can no longer be tied directly to a file.

Signed-off-by: Conrad Parker <conradparker@google.com>

### DIFF
--- a/common/inc/sgx_urts.h
+++ b/common/inc/sgx_urts.h
@@ -34,6 +34,8 @@
 #ifndef _SGX_URTS_H_
 #define _SGX_URTS_H_
 
+#include <stddef.h>
+
 #include "sgx_attributes.h"
 #include "sgx_error.h"
 #include "sgx_eid.h"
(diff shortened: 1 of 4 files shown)


ZongminGu:
Why is it important to load enclave from buffer?

birdperson-lives:
The use case we have in mind is loading an enclave from an ELF section in its loader's binary. There are also other conceivable use cases, such as loading an enclave received over a stream, or loading multiple enclaves from a memory-mapped file containing a sequence of enclaves.

lzha101:
We have implemented the feature by adding below API in the latest release. @birdperson-lives, I think we can close this PR now.
https://github.com/intel/linux-sgx/blob/e410bb922dd5bbba1d92a593b4a09c577b5de30c/psw/urts/linux/urts.cpp#L153

PR #326, opened 2018-10-08
Update sgx_read_rand.cpp

jasl:
Just format code for better reading

### DIFF
--- a/common/src/sgx_read_rand.cpp
+++ b/common/src/sgx_read_rand.cpp
@@ -45,7 +45,7 @@
 #define UINT32_MAX 0xFFFFFFFFU
 #endif
 
-static int g_is_rdrand_supported=-1;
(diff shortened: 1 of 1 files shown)


deeglaze:
Perhaps it's time this project got a .clang-format file and everything committed should be formatted with it.

jasl:
agree, which style you prefer to?

deeglaze:
I'm a Googler, so --style=google is good for me, but the Intel folks would need to agree on something. It looks like they prefer 4 space indentation instead of 2 space, for instance.

jasl:
Lol my partner is a ex-Googler, I think formatting whole project would be a big change, let Intel guys decide it

andyzyb:
Currently there is no plan to change the coding style which it was established a while ago.

Issue #330, opened 2018-10-17
Local attestation sample doesn't work in simulation mode


Joschuka:
Hello,

I followed every step in the installation description to install the Intel SGX SDK. Everything went well until I tried to execute the LocalAttestationExample, which compiles well but gives me the following error when trying to execute it :

`Load Enclave Failure
Available Enclaves
Enclave1 - EnclaveID 0
Enclave2 - EnclaveID 0
Enclave3 - EnclaveID 0Enclave1_test_create_session Ecall failed: Error code is 2002`

I tried to execute it as root but it doesn't work. Other examples have a similar error

intelphd:
Here's the output on the SealedData example :
`
	Replay Protected DRM operation:
cannot create enclave, error code = 0x3002
call create_sealed_policy fail, error code = 0x2002
Initialization the DRM failed.

	Replay Protected DRM update limitation:
cannot create enclave, error code = 0x3002
call create_sealed_policy fail, error code = 0x2002
Initialization the DRM failed.

	Replay Protected DRM replay attack protection:
cannot create enclave, error code = 0x3002
call create_sealed_policy fail, error code = 0x2002
Initialization the DRM failed.

	Time based policy operation:
cannot create enclave, error code = 0x3002
call create_time_based_policy fail, error code = 0x2002
Initialization the time based policy failed.

	Time based policy expiration:
cannot create enclave, error code = 0x3002
call create_time_based_policy fail, error code = 0x2002
Initialization the time based policy failed.
Enter a character before exit ...
`
I'm using the SDK on a fresh linux install, I have Ubuntu 18.04.

andyzyb:
The error shows loading Enclave failed.  Can you try to run the SampleEnclave project in simulation mode which is the simplest one?

andyzyb:
Would you please try the official build from 01.org and see if that works for you. 

https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/

intelphd:
It still doesn't work unfortunately

andyzyb:
What is the command you build local attestation sample?

intelphd:
I use the following one : `SGX_MODE=SW make`
It shows at the end : `The project has been built in debug simulation mode.`

llly:
This is a bug in our SGX SDK. Thanks for reporting.
Workaround is always setting launch_token to 0  when calling sgx_create_enclave() function.

llly:
[...]

PR #337, opened 2018-11-07
Fix uninitialized launch tokens, breaking simulation mode

fkirc:
This is a fix for issue #330.

In general, usage of uninitialized variables should be considered as a bug.

### DIFF
--- a/SampleCode/LocalAttestation/App/App.cpp
+++ b/SampleCode/LocalAttestation/App/App.cpp
@@ -73,7 +73,7 @@ uint32_t load_enclaves()
 {
     uint32_t enclave_temp_no;
     int ret, launch_token_updated;
-    sgx_launch_token_t launch_token;
+    sgx_launch_token_t launch_token = {0};
 
     enclave_temp_no = 0;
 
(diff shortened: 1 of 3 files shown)


lzha101:
Thanks for the patch. We have fixed this problem locally in another way.  It will be included in next upcoming release. Please stay tune.

fkirc:
Fixing it within the simulation code is fine, but these tests should not contain any usage of uninitialized variables.

lzha101:
This bug is fixed in the latest release. @fkirc, could you please try it and close this if no other problem?

fkirc:
Don't you think that launch tokens should be always initialized? (Regardless of whether the bug for simulation mode is fixed)

lzha101:
Agree that we need to initialize a variable before use it.  In the latest release, we allow to input NULL as the launch token and we've changed the sample code in that way. So no need to initialize launch token now.   https://github.com/intel/linux-sgx/blob/d3bcf9918da6575af5b8decfb29dedf1f55cca64/SampleCode/LocalAttestation/App/App.cpp#L79

fkirc:
Alright then I will close this request

PR #342, opened 2018-11-20
Bug fix in launch checker

dingelish:
The current launch checker misses an essential check on launch token (I guess it's commented out for debug). This forces launch token always updated and results in the inconsistency between launch token design and implementation.


lzha101:
The launch token updates logic is moved to enclave_common module  to support Microsoft Open Enclave project.  See https://github.com/intel/linux-sgx/blob/master/psw/enclave_common/sgx_enclave_common.cpp#L415 
Current version has some legacy code/implementation about launch token.  This will be fixed in the next  upcoming release. Thanks.

PR #346, opened 2018-12-02
Adding error messages for debug purposes

Icaro-Lima:
```
    /* Step 2: call sgx_create_enclave to initialize an enclave instance */
    /* Debug Support: set 2nd parameter to 1 */

    ret = sgx_create_enclave(TEST_FILENAME, SGX_DEBUG_FLAG, &token, &updated, &global_eid, NULL);

    printf("---> %s\n", sgx_create_enclave_err_msg(ret));
```
```
token_path: /home/icarolima/enclave.token
---> The enclave was loaded and initialized successfully.
IN TEST
Application ran with success
```


Icaro-Lima:
I will rename the method.

andyzyb:
This patch is unlikely to be accepted since it is not clearly stating which problem to solve.  If it is just to convert an error code to a string, then it doesn't seem provide much value.

Icaro-Lima:
I disagree, debug messages are one of the major reason for extensive code at least in the ones I write and in the sample codes from this repository. I'll edit the merge request message to make it clearer.

pdxjohnny:
I also think that this is useful. I did something similar and then went and looked at PRs and saw this. I think its important to have some function like this or the one I added (which is admittedly very basic). But some way of quickly doing an error code -> string of the message or string of the description. Preferably both

PR #366, opened 2019-01-23
headers: Add macro to print error code as string

johnandersen777:
* Adds a `PRINT_SGX_STATUS` macro to sgx_error.h which prints
  the `sgx_status_t` as a string.

### DIFF
--- a/common/inc/sgx_error.h
+++ b/common/inc/sgx_error.h
@@ -117,4 +117,200 @@ typedef enum _status_t
 
 } sgx_status_t;
 
+#define PRINT_SGX_STATUS(FILE_STRUCT, STATUS_T) \
+    switch (STATUS_T) {\
+    case SGX_SUCCESS:\
+        fprintf(FILE_STRUCT, "Error: SGX_SUCCESS\n");\
+        break;\
+    case SGX_ERROR_UNEXPECTED:\
+        fprintf(FILE_STRUCT, "Error: SGX_ERROR_UNEXPECTED\n");\
+        break;\
+    case SGX_ERROR_INVALID_PARAMETER:\
+        fprintf(FILE_STRUCT, "Error: SGX_ERROR_INVALID_PARAMETER\n");\
+        break;\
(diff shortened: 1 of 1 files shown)


pdxjohnny:
Generated with:

```console
(
echo -e '#define PRINT_SGX_STATUS(FILE_STRUCT, STATUS_T) \\\n    switch (STATUS_T) {\\';
for error in $(grep -E '=.*SGX_MK_ERROR' /opt/intel/sgxsdk/include/sgx_error.h | awk '{print $1}' | sed 's/=//g'); do
  printf '    case %s:\\\n        fprintf(output, "Error: %s\\n");\\\n        break;\\\n' $error $error;
done;
echo -e '    }'
)
```

pdxjohnny:
maybe this would be better as something similar to `perror` called `sgx_perror` or something. Where it acts just as `perror` does but for `sgx_status_t` errors

sergefdrv:
Might be even better to make it similar to `strerror`, for fine control of output. Think of advanced logging mechanism.

pdxjohnny:
Ya that is probably a better option, I think that's what was started in #346 perhaps that PR should be one PR introducing the `strerror` and another adding the logging to various places.

pdxjohnny:
Something similar is going on in the examples

https://github.com/intel/linux-sgx/blob/a85fbe9a55767681b0f799a66555b123312fb72c/SampleCode/SampleEnclave/App/App.cpp#L55-L70

Icaro-Lima:
Hello, I was disinclined to continue working on my pull request after the comment from the Intel member himself saying that maybe it did not help much. And I still maintain my position, I think it is very important and avoids many lines of unnecessary codes that will be inserted in each enclave.

pdxjohnny:
I agree. As I posted above something similar was done in the sample enclave code. If its useful there then it is of course useful and instead of duplicating that functionality it should be a part of the SDK.

LuminousXLB:
I believe that it's a really useful patch.
FYI, here's how I did the same thing.

https://pastebin.com/PT4Ds8WQ

PR #368, opened 2019-02-05
Remove unused ill-defined 'storage' type field

deeglaze:
The templated "storage" field in randomly_placed_buffer is uses alignas
in an undefined way. The field is also not used, so this change removes
it.

### DIFF
--- a/common/inc/sgx_random_buffers.h
+++ b/common/inc/sgx_random_buffers.h
@@ -249,9 +249,6 @@ struct alignas(A)randomly_placed_buffer
         free(ptr);
     }
 
-    template <unsigned C = 1>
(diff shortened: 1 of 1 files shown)


lzha101:
Do you encounter any error with the 'storage' field?

deeglaze:
> Do you encounter any error with the 'storage' field?

linux_sgx/common/inc/sgx_random_buffers.h:254:35: error: 'alignas' attribute cannot be applied to types
    using storage = char[size(C)] alignas(A);

ZongminGu:
What is the compiler you are using?

deeglaze:
This error is produced by clang, which I know you don't support. This PR is addressing unspecified behavior, not specifically clang support.

PR #369, opened 2019-02-05
Do not mix static and extern "C"

deeglaze:
The meaning of 'static' is to make a symbol private to the source
file, whereas 'extern "C"' is meant for linking symbols across
files with an unmangled name. The combination of the two allowed
by GCC is non-standard.

### DIFF
--- a/sdk/trts/trts_switchless.cpp
+++ b/sdk/trts/trts_switchless.cpp
@@ -35,11 +35,9 @@
 #include "global_data.h"
 #include "util.h"
 
-extern "C" {
-
 struct sl_uswitchless;
 
-static sgx_status_t dummy_init_switchless(void* ms) {
+extern "C" sgx_status_t dummy_init_switchless(void* ms) {
     UNUSED(ms);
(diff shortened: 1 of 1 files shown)


sdk/trts/trts_switchless.cpp
```diff
@@ -35,23 +35,19 @@
 #include "global_data.h"
 #include "util.h"
 
-extern "C" {
-
 struct sl_uswitchless;
```
lzha101: This line seems unnecessary and could be removed.

lzha101:
And please add your Signed-off-by in the commit message before we merge the patch.

sdk/trts/trts_switchless.cpp
deeglaze: The struct line doesn't affect the extern "C" / static selection. I'd prefer to keep that change separate.

deeglaze:
> And please add your Signed-off-by in the commit message before we merge the patch.

Done.

PR #391, opened 2019-04-22
Download dcap from the archive if no git repository is found

### LINKED ISSUE #390: SGX 2.5 release `make` fails due to git module command
From the `Makefile` of SDK 2.5,
```
all: dcap_source sdk psw

# dcap_source:
# 	git submodule update --init --recursive
```
Simply running `make` from the source code downloaded from [here](https://github.com/intel/linux-sgx/releases/tag/sgx_2.5) won't work due to `git submodule` command

I had to manually download the source code for DataCenterAttestationPrimitives.

IMHO It would be good if the release source tar bells download source codes from the archive.

shpark:
This closes #390

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ clean:
 	@$(MAKE) -C sdk/                                clean
 	@$(MAKE) -C psw/                                clean
 	@$(RM)   -r $(ROOT_DIR)/build
-	@$(RM)   -rf external/dcap_source
+	@$(RM)   -rf external/dcap_source/*
 	@$(RM)   -r linux/installer/bin/sgx_linux*.bin
 	@$(RM)   -r linux/installer/deb/libsgx-enclave-common/libsgx-enclave-common-dbgsym_*
 	@$(RM)   -r linux/installer/deb/libsgx-enclave-common/libsgx-enclave-common_*.tar.*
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -68,6 +78,7 @@ clean:
 	@$(MAKE) -C sdk/                                clean
 	@$(MAKE) -C psw/                                clean
 	@$(RM)   -r $(ROOT_DIR)/build
+	@$(RM)   -rf external/dcap_source
```
llly: `external/dcap_source` folder is part of the repo. Unstaged changes exist after `make clean`. Only remove  the contents inside the folder is enough.

shpark: I just removed this line. Simply calling `rm -rf external/dcap_source/*` leaves `.git` behinds. Maybe we need something like `make -C external/dcap_source clean`?

Makefile
```diff
@@ -36,7 +38,14 @@ include buildenv.mk
 all: dcap_source sdk psw
 
 dcap_source:
+ifeq ($(shell git rev-parse --is-inside-work-tree), true)
 	git submodule update --init --recursive
+else
+	curl --output dcap_source.tar.gz -L --tlsv1 ${DCAP_DOWNLOAD_BASE}/DCAP_${DCAP_VER}.tar.gz
```
thiagomacieira: DO NOT download during Make.

Please add this to the script that downloads dependencies prior to build. instead.

shpark: Sorry for the late response, I've missed to answer to your comment. Then, should this change be reverted?

PR #396, opened 2019-05-02
Fix a potential expression issue.

dingelish:
Hi there, I find a potential expression issue coincidentally. The diff is clear.

The removed expressions contain comparisons between `size_t  smax`/`size_t n` and `SIZE_MAX` which is defined in common/inc/tlibc/sys/stdint.h as:
```
131 #ifdef __x86_64__
132 #define UINT64_MAX      0xffffffffffffffffUL
133 #else
...
180 #ifdef __x86_64__
181 #define INTPTR_MIN      INT64_MIN
182 #define INTPTR_MAX      INT64_MAX
183 #define UINTPTR_MAX     UINT64_MAX
...
190 /* 7.18.2.5 Limits of greatest-width integer types */
191 #define INTMAX_MIN      INT64_MIN
192 #define INTMAX_MAX      INT64_MAX
193 #define UINTMAX_MAX     UINT64_MAX
...
207 /* Limits of size_t (also in limits.h) */
208 #ifndef SIZE_MAX
209 #define SIZE_MAX        UINTPTR_MAX
210 #endif
```

I think `SIZE_MAX` would finally be defined as `0xffffffffffffffffUL`. As a result, these two comparison expressions are meaningless. This kind of code is categorized as [CWE-569](http://cwe.mitre.org/data/definitions/569.html) and could be found by Coverity checker `CONSTANT_EXPRESSION_RESULT`.

I find this coincidentally by using `cargo clippy` to examine my ported Rust-version sgx_crypto_helper. Not sure if it's a real bug. Please advise. Thanks!


dingelish:
Further investigation on this code reveals that this `memset_s` seems likely to be "ISO/IEC 9899:2011 section K.3.7.4.1 The memset_s function", proposed [here](https://mail-index.netbsd.org/tech-userlevel/2012/02/24/msg006125.html). It refers to type `[rsize_t](https://en.cppreference.com/w/c/error)` which is:

> a typedef for the same type as size_t, used to self-document functions that range-check their parameters at runtime
And `RSIZE_MAX`
> largest acceptable size for bounds-checked functions, expands to either constant or variable which may change at runtime (e.g. as the currently allocated memory size changes)

Should we choose a proper `RSIZE_MAX` here to make this `memset_s` works as expected?

PR #399, opened 2019-05-10
Solve unknown type name ‘bool’ when linking sgx_uae_service.h

lanhaoxiang:


llly:
sgx_uae_service.h doesn't use `bool` anymore in sgx_2.6.

PR #401, opened 2019-05-14
Select proper kernel driver

garywest:
Selects the proper kernel driver.


ZongminGu:
My personal opinion is: If user has two drivers on a machine. User should take the responsibility to modprobe the one he/she want to use, instead of the install script. So seems that the current install script is not correct too, the script should not modprobe any driver.

PR #450, opened 2019-10-01
Add missing #include to trts_veh.cpp

deeglaze:
The ce_cdefs header defines SE_64, which this source file depends on.
When building with clang, this missing header causes the build to break.

Signed-off-by: Dionna Glaze <drdeeglaze@gmail.com>


andyzyb:
Thanks for the patch. We'll merge it later.

PR #459, opened 2019-10-21
Add missing mkdir -p /var/opt/aesmd to aesmd.service

exFalso:
This directory was added in 2.7, but it isn't mkdir-d in the service file (probably relies on the startup.sh hook)


xiangquanliu:
“/var/opt/aesmd” will be created by installer. And it should be always there since it is not tmpfs. So there is no need to add this line.

llly:
`/var/run/aesmd/` need to be created in `aesmd.service` because `/run` is tmpfs and the folder doesn't exist after each OS boot, the files in it are created by `aesm_service`. `/var/opt/aesmd/` always exists and initial files in it are installed by `startup.sh`.

exFalso:
This is to enable installations which don't use `startup.sh` (in particular I repackaged aesmd for NixOS), but use the service file directly instead. Thought it may be useful, but it's not necessary (have a workaround for it)

PR #476, opened 2019-12-16
Use ByteSizeLong() instead of deprecated ByteSize()

siburu:
Since protobuf 3.11.0, deprecated codes are annotated and warned.
https://github.com/protocolbuffers/protobuf/pull/6612

"make psw" fails because of uses of deprecated method ByteSize().
This PR simply replaces all calls to ByteSize() with ByteSizeLong().


llly:
protobuf 3.11.0 and later is incompatible with early releases. 
We can not change these functions because we must support the default protobuf provided by the OSes referred to in Readme.
Best solution is changing CXXFLAGS to disable the warings and it works for all protobuf version.

siburu:
Thank you! Now I understand the situation.
I will change this PR to add "-wno-deprecated" flag for protobuf-related stuffs.

llly:
We plan to fix this compatibility issue in next major release.

siburu:
Thanks. Okay, is it better to just wait for it? Is the release relatively close?

llly:
probably in early Jan. 2020.

siburu:
Very close. Thank you!!
I will wait for it following your suggestion to suppress warnings.

PR #500, opened 2020-02-22
Update Makefile for RemoteAttestation

tygert:
This updates the Makefile for SampleCode/RemoteAttestation in order to incorporate the current directory structure of the repo. My changes work, but you may want to fix things some other way. At any rate, the current version of your repo for SampleCode/RemoteAttestation does not work with your present directory structure.

### DIFF
--- a/SampleCode/RemoteAttestation/Makefile
+++ b/SampleCode/RemoteAttestation/Makefile
@@ -103,7 +103,7 @@ else
 endif
 
 App_Cpp_Flags := $(App_C_Flags)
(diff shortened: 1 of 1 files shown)


lzha101:
The existing Makefile in RemoteAttestation sample is designed to work with SGX SDK installer only. If you have installed the SGX SDK installer, you could find a version of RemoteAttestation sample under ${sgx-sdk-install-path}/SampleCode/RemoteAttestation/. Please build that copy rather than directly build from the source repo folder.

tygert:
Thanks, @lzha101 -- that does work fine. You might want to update the documentation (README.txt) to mention that you do not support building in the source repo except for the initial installation -- RemoteAttestation is the only example in SampleCode which does not work when built in the source repo, and not being able to run the Makefile along with the source codes in the source repo might seem a bit strange without accompanying documentation explaining otherwise. That the installed version exists and differs from the version in the source repo is not totally obvious.

PR #506, opened 2020-02-28
Change 'is not' to '!=' in gdb_sgx_plugin.py

lkatalin:
This change would make the script compatible with Python3.


xxu36:
I think it is correct. We should use ==/!= to compare the value and is/is not to compare the object identity. Thanks!

lzha101:
Could you refer to CONTRIBUTING.md to add your signed-off-by in the commit message?

lkatalin:
@lzha101 I added the signed-off-by. Sorry I missed that!

PR #512, opened 2020-03-19
config: aesmd.service support for new device files

ipuustin:
The current upstream candidate driver users `/dev/sgx/enclave` and `/dev/sgx/provision` files. Add the device files to aesmd systemd service file so that they can be accessed by the service.


andyzyb:
Thanks.  It makes sense.  We'll test it before merging.

PR #513, opened 2020-03-19
Malloc Shim

michaelkaplan13:


andyzyb:
Thanks for the PR.  But it is unlikely to be accepted.  malloc is a weak symbol, and can be overwritten by your own malloc implementation.  There is no need to wrap it again.   

 32: 0000000000000000  6678 FUNC    WEAK   HIDDEN     8 malloc

michaelkaplan13:
Sorry, I didn't actually mean to make this PR in the Intel repo. We were tinkering around with our own internal fork of the repo, and I just made the PR to show someone else the changes I had made. Thanks for the response nonetheless!

PR #515, opened 2020-03-22
Fix aesmd not starting if /dev is mounted as noexec

assafmo:
See https://lore.kernel.org/linux-sgx/4db41057-910c-b686-0428-474debe382c1@fortanix.com/

### DIFF
--- a/psw/ae/aesm_service/config/aesmd_service/aesmd.service
+++ b/psw/ae/aesm_service/config/aesmd_service/aesmd.service
@@ -10,6 +10,7 @@ Environment=AESM_PATH=@aesm_folder@
 Environment=LD_LIBRARY_PATH=@aesm_folder@
 WorkingDirectory=@aesm_folder@
 PermissionsStartOnly=true
+ExecStartPre=/bin/mount -o remount,exec /dev
 ExecStartPre=@aesm_folder@/linksgx.sh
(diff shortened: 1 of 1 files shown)


assafmo:
So actually this didn't work for me. Instead of using `ExecStartPre` I had to create a startup script especially for remounting `/dev` as `exec`.

I don't know what you'd want to do with this PR or even if you want to support ubuntu != {16.04,18.04}.

Anyway I think the driver should make sure it is mounted as `exec`.

See https://github.com/enigmampc/EnigmaBlockchain/commit/866949a926edc9549c07f53545bcf13f97faef25?diff=split#diff-fc708eba6ce617e52f528df7404127e6R77-R90

andyzyb:
I somewhat agree that it is better to leave this to driver.  In some cases there is no AESM installed on the system, then user application needs to handle this?

assafmo:
I agree

assafmo:
I don't this the user application needs to handle this, but the driver. I wouldn't want my apps to remount `/dev` on my system.  
Though if not AESM then I think the driver should handle this.

haitaohuang:
This is discussed in LKML for upstreaming driver. It needs be a solution from driver

kaimast:
Thanks for posting the fix @assafmo!

PR #522, opened 2020-04-06
Update SDK installer to provide sim libs

dcmiddle:
Simulation libraries are provided through the SDK installation into
the sdk_libs directory and the environment file adds that directory to
LD_LIBRARY_PATH. Starting in version 2.8 new simulation libraries were
introduced but not added to this path.

This commit adds libsgx_quote_ex_sim.so, libsgx_epid_sim.so, and
libsgx_launch_sim.so to the expected directory.

Without this change applications following the conventions of the sample
Makefiles may build but fail to execute.

Signed-off-by: Dan Middleton <dan.middleton@intel.com>


llly:
Reasonable.Will merge it soon.

PR #525, opened 2020-04-09
Can't build SGX SDK with the current dockerfile

jbdelcuv:
Change the make target to build the SDK with no mitigation to avoid having to install the binutils package that contains the updated assembler.


andyzyb:
This is not an officially supported build configuration for SDK and it is not well tested.

PR #526, opened 2020-04-14
Fix sgx-aesm-service dh_shlibdeps rules

dingelish:
`dh_shlibdeps` should always accompanied by `--dpkg-shlibdeps-params=--ignore-missing-info`. If not, `make deb_psw_pkg` would fail on some platforms where libstdc++ is not installed from software package. All other two appearances of `dh_shlibdeps` are both correct.

```
$ grep -R override_dh_shlibdeps -A3
linux/installer/deb/libsgx-urts/libsgx-urts-1.0/debian/rules:override_dh_shlibdeps:
linux/installer/deb/libsgx-urts/libsgx-urts-1.0/debian/rules-	dh_shlibdeps -l$(LINUX_BUILD_DIR) --dpkg-shlibdeps-params=--ignore-missing-info
--
linux/installer/deb/sgx-aesm-service/sgx-aesm-service-1.0/debian/rules:override_dh_shlibdeps:
linux/installer/deb/sgx-aesm-service/sgx-aesm-service-1.0/debian/rules-	dh_shlibdeps -l $(LINUX_BUILD_DIR)
--
linux/installer/deb/libsgx-enclave-common/libsgx-enclave-common-1.0/debian/rules:override_dh_shlibdeps:
linux/installer/deb/libsgx-enclave-common/libsgx-enclave-common-1.0/debian/rules-	dh_shlibdeps -l $(LINUX_BUILD_DIR) --dpkg-shlibdeps-params=--ignore-missing-info
```

### DIFF
--- a/linux/installer/deb/libsgx-epid/libsgx-epid-1.0/debian/rules
+++ b/linux/installer/deb/libsgx-epid/libsgx-epid-1.0/debian/rules
@@ -4,5 +4,8 @@
 # Uncomment this to turn on verbose mode.
 #export DH_VERBOSE=1
 
+override_dh_shlibdeps:
+	dh_shlibdeps --dpkg-shlibdeps-params=--ignore-missing-info
+
 %:
 	dh $@
--- a/linux/installer/deb/libsgx-launch/libsgx-launch-1.0/debian/rules
+++ b/linux/installer/deb/libsgx-launch/libsgx-launch-1.0/debian/rules
@@ -4,5 +4,8 @@
 # Uncomment this to turn on verbose mode.
 #export DH_VERBOSE=1
 
+override_dh_shlibdeps:
+	dh_shlibdeps --dpkg-shlibdeps-params=--ignore-missing-info
+
 %:
 	dh $@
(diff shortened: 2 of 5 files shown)


xiangquanliu:
The option --ignore-missing-info is discouraged to use from the dpkg-shlibdeps manual. 
--ignore-missing-info
              Do not fail if dependency information can't be found for a shared library (since dpkg 1.14.8).  Usage of this option is discouraged, all libraries should provide dependency information (either with shlibs files, or with symbols files) even if they are not yet used by other packages.

Moreover, package libstdc++ should be installed as dependency when installing the needed packages from the README.

dingelish:
as of linux-sgx 2.9.1, the required gcc/binutils version is extremely high (>8.4, >9.3). seldom linux distribution provides them. if we can't make linux-sgx depends on in-house gcc/binutils/glibc, we will be exposed to the vulnerabilities for a long time, until the linux distribution provides them. i don't think it's a good choice.

xiangquanliu:
So what OSes do you use to build the debian packages?

dingelish:
I'm using Debian 8 and Debian 9. Some of my friends are using [CentOS](https://github.com/apache/incubator-teaclave-sgx-sdk/issues/158)

xiangquanliu:
Right now there is no plan to support Debian OS.

andyzyb:
As each Linux distro has slight difference it would be difficult to support all of them.  It is welcome for any patch to enable other distros, but the change is better to be isolated and has no impact to other distros.

PR #530, opened 2020-04-23
Changes for kernel sgx patch V29

haitaohuang:

### DIFF
--- a/psw/enclave_common/sgx_enclave_common.cpp
+++ b/psw/enclave_common/sgx_enclave_common.cpp
@@ -389,7 +389,7 @@ extern "C" void* COMM_API enclave_create(
         hdevice_temp = s_hdevice;
     }
     
-    void* enclave_base = mmap(base_address, enclave_size, PROT_NONE, MAP_SHARED, hdevice_temp, 0);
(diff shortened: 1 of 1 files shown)


psw/enclave_common/sgx_enclave_common.cpp
```diff
@@ -742,6 +749,18 @@ extern "C" bool COMM_API enclave_initialize(
             *enclave_error = ENCLAVE_INVALID_PARAMETER;
         return false;
     }
+    int hfile = -1;
+    if (s_driver_type == SGX_DRIVER_IN_KERNEL)
+    {
+        hfile = get_file_handle_from_address(base_address);
+        if (hfile == -1)
+        {
+        SE_TRACE(SE_TRACE_WARNING, "\nSGX_IOC_ENCLAVE_INIT failed - %p is not a valid enclave \n", base_address);
```
JunjunGu1, line 758: indent is not correct here

haitaohuang: thanks. Fixed

haitaohuang:
Based on LKML discussion I changed the initial mmap for enclave address range to MAP_PRIVATE from MAP_SHARED. The later mmap calls did not change.

Tested with SampleCommonLoader and SampleEnclaves, still works for both v29 and v28

PR #542, opened 2020-05-12
loading enclave will fail in enclave_create with oot driver

JunjunGu1:
https://github.com/intel/linux-sgx/pull/530 doesn't consider the oot driver.
Loading enclave will fail in enclave_create with oot driver.
This patch will fix this problem.

Signed-off-by: Gu, Junjun junjun.gu@intel.com

### DIFF
--- a/psw/enclave_common/sgx_enclave_common.cpp
+++ b/psw/enclave_common/sgx_enclave_common.cpp
@@ -392,7 +392,7 @@ extern "C" void* COMM_API enclave_create(
     
     if(s_driver_type == SGX_DRIVER_IN_KERNEL)
     {
-        enclave_base = mmap(base_address, enclave_size, PROT_NONE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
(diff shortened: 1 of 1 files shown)


psw/enclave_common/sgx_enclave_common.cpp
```diff
@@ -389,7 +390,15 @@ extern "C" void* COMM_API enclave_create(
         hdevice_temp = s_hdevice;
     }
     
-    void* enclave_base = mmap(base_address, enclave_size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    if(s_driver_type == SGX_DRIVER_IN_KERNEL)
+    {
+        enclave_base = mmap(base_address, enclave_size, PROT_NONE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
```
haitaohuang, line 395: Need change to MAP_PRIVATE so larger address space can be mapped

JunjunGu1: thanks. fixed.

PR #557, opened 2020-06-23
Fix an error in docker/build

haitaohuang:
And minor README update


jbdelcuv:
Looks good to me.

PR #561, opened 2020-07-01
Support to dynamically link with libsgx_launch.so

jiazhang0:
When loading a sgxsdk-written library with dlopen, the libsgx_launch.so will not be loaded immediately until enclave creation function sgx_create_enclave() is called. Before that,
current process may enter into a container before calling sgx_create_enclave(). If so,
libsgx_launch.so has to exist in container, but this is not always true. The following diagram
shows what happens:

```
dlopen("sgxsdk_based.so", RTLD_NOW) and foo = dlsym("foo")
 |
\|/
unshare()
 |
\|/
(*foo)()
 |
+-> sgx_create_enclave()
    |
    +-> dlopen("libsgx_launch.so")
```

The solution is to implement dynamic link with libsgx_launch.so. So using dlopen() to load
libsgx_launch.so should be dropped.

Signed-off-by: Jia Zhang <zhang.jia@linux.alibaba.com>

### DIFF
--- a/psw/ae/aesm_service/source/bundles/le_launch_service_bundle/le_launch_service_bundle.cpp
+++ b/psw/ae/aesm_service/source/bundles/le_launch_service_bundle/le_launch_service_bundle.cpp
@@ -21,8 +21,6 @@ extern "C" bool is_launch_token_required();
 extern ae_error_t start_white_list_thread(unsigned long timeout=THREAD_TIMEOUT);
 extern ThreadStatus white_list_thread;
 
-extern "C" void init_get_launch_token(const func_get_launch_token_t func);
-
 extern "C" sgx_status_t get_launch_token(const enclave_css_t *signature,
                                          const sgx_attributes_t *attribute,
                                          sgx_launch_token_t *launch_token)
@@ -174,7 +172,6 @@ private:
         auto context = cppmicroservices::GetBundleContext();
         get_service_wrapper(g_network_service, context);
         get_service_wrapper(g_launch_service, context);
-        init_get_launch_token(::get_launch_token);
         start_white_list_thread(0);
         initialized = true;
         AESM_DBG_INFO("le bundle started");
--- a/psw/enclave_common/Makefile
+++ b/psw/enclave_common/Makefile
@@ -51,13 +51,13 @@ INC += -I$(SGX_HEADER_DIR)                \
        -I$(LINUX_PSW_DIR)/urts/linux      \
        -I$(LINUX_PSW_DIR)/enclave_common
 
-LDFLAGS := -lwrapper -ldl
+LDFLAGS := -lwrapper -lsgx_launch
 LDFLAGS += $(COMMON_LDFLAGS) -Wl,--version-script=sgx_enclave_common.lds -Wl,--gc-sections
 
 DIR1 := $(LINUX_PSW_DIR)/enclave_common
 DIR2 := $(LINUX_PSW_DIR)/urts/linux
 
-LIB += -L$(COMMON_DIR)/se_wrapper_psw
+LIB += -L$(COMMON_DIR)/se_wrapper_psw -L$(LINUX_PSW_DIR)/uae_service/linux
 
 OBJ := sgx_enclave_common.o edmm_utility.o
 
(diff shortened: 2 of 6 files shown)


llly:
The logic to dlopen("libsgx_launch.so") is designed on purpose.
When using DCAP platform, `libsgx_enclave_common.so` won't call get_launch_token() from `libsgx_launch.so`. `libsgx_enclave_common.so` need to be work without `libsgx_launch.so`.

andyzyb:
I think the problem of this PR is not very clear.  It still needs to call get_launch_token in enclave_initialize with this PR, how can sgx_launch.so not exist in the container if application enters container?

jiazhang0:
@llly So is it possible to use this way to get DCAP platform working?
```C
        if (0 == enclave_css->header.hw_version && s_driver_type != SGX_DRIVER_DCAP) {
            sgx_status_t status = get_launch_token(enclave_css, &it->second, &launch_token);
            if (status != SGX_SUCCESS) {
                if (enclave_error != NULL)
                    *enclave_error = error_aesm2api(status);
                return false;
            }
        }
```

jiazhang0:
@andyzyb Before entering into the container, the current process will dlopen() with RTLD_NOW flag to ensure liblibsgx_enclave_common.so, libsgx_urts.so and the dependent libsgx_launch.so are completely loaded from host side.

After entering into the container, calling get_launch_token in enclave_initialize will be handled inside current process without the assist from dynamic linker in container, e.g, imaging that the current process running on a centos and the container's os is alpine, which has a completely different dynamical linker and libc such as musl.

jiazhang0:
Update the commit header for the calling diagram.

haitaohuang:
Not sure the exact use case here. I wonder why you can't mount the libs into container if the libs are needed for app to work....
Besides, you will have to fix all libs that do dlopen. So this won't be a generic solution.

jiazhang0:
> Not sure the exact use case here. I wonder why you can't mount the libs into container if the libs are needed for app to work....
> Besides, you will have to fix all libs that do dlopen. So this won't be a generic solution.

The use case is `rune`, which is provided by https://github.com/alibaba/inclavare-containers and implements the protected container runtime for confidential computing.

`rune` needs the so-called enclave runtime PAL (in the form of .so) to interface with libos six, e.g, occlum and graphene. Taking occlum as example, pal_init_enclave (https://github.com/occlum/occlum/blob/master/src/pal/src/pal_enclave.c#L59) is the subroutine of enclave runtime PAL API pal_init() defined by the spec (https://github.com/alibaba/inclavare-containers/blob/master/rune/libenclave/internal/runtime/pal/spec_v2.md#2pal_init). pal_init_enclave calls intel sgxsdk API (https://github.com/occlum/occlum/blob/master/src/pal/src/pal_enclave.c#L105) to load signed enclave image.

From the perspective of `rune`, it loads enclave runtime PAL before entering into container with dlopen(RTLD_NOW), and calls pal_init and the further pal_init_enclave after entering into container. In other word, the load operation and execution operation are separate and across container boundary.

If libsgx_launch.so is provided in container, this is also a potential security issues, because the host outside container is usually more privileged. You can refer to this link for the details:

- https://github.com/moby/moby/pull/39612/files/a316b10dab79d9298b02c7930958ed52e0ccf4e4

So `rune` needs to make sure the enclave runtime PAL and its dependent libraries are loaded from host side. This limits the flexibility of PAL implementation but it is not a big deal.

andyzyb:
@jiazhang0  thanks for explaining the usage.  This is interesting as it loads all the dependent libraries from host OS then moves to container for execution.   We probably cannot change the current logic as libsgx_launch.so is not always required/installed on FLC enabled system, let libsgx_enclave_com.so link to libsgx_launch.so in build time seems not reasonable.

andyzyb:
Have you tried to dlopen libsgx_launch.so in the host environment for rune, then in the container even it is called again it will not actually take effect?

jiazhang0:
> Have you tried to dlopen libsgx_launch.so in the host environment for rune, then in the container even it is called again it will not actually take effect?
> 
> https://stackoverflow.com/questions/12148454/will-dlopen-yield-the-same-handle-for-two-calls-with-the-same-file

This workaround works. If libsgx_launch.so is loaded on host side, then calling dlopen it in container would not introduce another load. However, introducing such a workaround in rune looks like not a general solution, because rune is not designed for sgxsdk. It supports intel sgx technology but it is not use intel sgxsdk at all.

andyzyb:
I agree.  But at lease this  can be a temporary solution without accepting this PR.

jiazhang0:
> I agree. But at lease this can be a temporary solution without accepting this PR.

OK. I will work around this issue in rune side.

PR #582, opened 2020-08-07
docker: use volumes for sharing the aesmd socket.

ipuustin:
When sharing `aesmd` socket between two docker containers, use a dedicated volume instead of temporary directory on the host.

### DIFF
--- a/docker/build/README.md
+++ b/docker/build/README.md
@@ -34,7 +34,7 @@ The Dockerfile specifies 3 image build targets:
 2. aesm: Takes the PSW installer from builder to install and run the AESM daemon.
 3. sample: Installs the SDK installer from builder, then builds and runs the SampleEnclave app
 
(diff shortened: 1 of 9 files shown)


ipuustin:
Do you mean that something else is expecting the `/tmp/aesmd` directory to be present? Having `/var/run/aesmd` mounted in a volume should be transparent to the prebuilt aesmd putting the socket there.

haitaohuang:
We have another docker file here: https://github.com/intel/linux-sgx/tree/master/linux/installer/docker while shows how to use prebuilt PSW and SDK. Can you also update scripts in that directory?

haitaohuang:
Looks good to me

lzha101 (approved):

haitaohuang:
@ipuustin 
After some testing, I think the volume needs be mount with rw flag and created as following:
"docker volume create --driver local --opt type=tmpfs --opt device=tmpfs --opt o=rw aesmd-socket"
Could you update the changes?
If we run aesmd as root it would probably be fine. But let's address that issue separately.

ipuustin:
Thanks for the review, I updated the patch now.

PR #583, opened 2020-08-07
docker: do not run aesmd as user "aesmd".

ipuustin:
When the devices are mounted to container using `--device` Docker command line option, they are readable/writable only by root. Running aesmd as user "aesmd" leads to a "permission denied" error when accessing `/dev/sgx/enclave`.

Also, for the same reason, run the sample app as root.

### DIFF
--- a/docker/build/Dockerfile
+++ b/docker/build/Dockerfile
@@ -75,7 +75,6 @@ RUN apt-get update && apt-get install -y \
 WORKDIR /installer
 COPY --from=builder /linux-sgx/linux/installer/bin/*.bin ./
 RUN ./sgx_linux_x64_psw*.bin --no-start-aesm
-USER aesmd
 WORKDIR /opt/intel/sgxpsw/aesm/
 ENV LD_LIBRARY_PATH=.
 CMD ./aesm_service --no-daemon
@@ -97,9 +96,6 @@ RUN sh -c 'echo yes | ./sgx_linux_x64_sdk_*.bin'
 
 WORKDIR /opt/intel/sgxsdk/SampleCode/SampleEnclave
 RUN SGX_DEBUG=0 SGX_MODE=HW SGX_PRERELEASE=1 make
-
-RUN adduser -q --disabled-password --gecos "" --no-create-home sgxuser
-USER sgxuser
 
 CMD ./app
 


haitaohuang:
It is working on my Ubuntu 18.04 box and both ubuntu and debian base images. Is this env/OS specific? 

My docker version:
Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:44:07 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

ipuustin:
Hmm, could be, or maybe the new in-tree driver sets tighter permissions? I only tested this on Clear Linux. Can you tell me which permissions `/dev/sgx/enclave` has in your Docker environment and what `whoami` returns there?

haitaohuang:
sgxuser@569d101d8bd8:/$ whoami
sgxuser
sgxuser@569d101d8bd8:/$ ls /dev/sgx -al
total 0
drwxr-xr-x 2 root root     60 Aug 19 16:12 .
drwxr-xr-x 6 root root    380 Aug 19 16:12 ..
crw-rw-rw- 1 root root 10, 56 Aug 19 16:12 enclave

sorry for delay

haitaohuang:
on aesmd docker: 
aesmd@9adb39e7803e:/opt/intel/sgx-aesm-service/aesm$ whoami
aesmd
aesmd@9adb39e7803e:/opt/intel/sgx-aesm-service/aesm$ ls -al /dev/sgx
total 0
drwxr-xr-x 2 root root     80 Aug 19 16:33 .
drwxr-xr-x 6 root root    400 Aug 19 16:33 ..
crw-rw-rw- 1 root root 10, 56 Aug 19 16:33 enclave
crw-rw---- 1 root 1001 10, 55 Aug 19 16:33 provision

ipuustin:
For comparison, this is what I see in my container env:
```
# ls -ld /dev/sgx/*
crw------- 1 root root 10, 62 Aug 13 07:27 /dev/sgx/enclave
crw------- 1 root root 10, 61 Aug 13 07:27 /dev/sgx/provision
```

The permissions are the same also outside of the container. How are your permissions outside of the container?

It looks like in your environment `/dev/sgx/enclave` is writable by `other`, which means that in your environment the enclaves can be created. Did you try if remote attestation works, because only `group` can write to `/dev/sgx/provision`? Or is user `aesmd` defined to be member of group `1001` in your container?

haitaohuang:
After some more testing on my side. I agree the easiest solution is to run aesmd as root for the provisioning access or we find a better way to map host group/permission to container group/permission. We'll run aesmd as root for now

With udev rule to set world RW for /dev/sgx/enclave, do you see need to run apps as root?

PR #592, opened 2020-09-07
Fixed minor errata

Zildj1an:
That dot shouldn't be there, otherwise git will attempt to use "." as the destination path, which obviously already exists.
Hence, if you follow the tutorial you get:
```
~/Desktop/SGX[22:14:59]$ git clone https://github.com/intel/linux-sgx.git .
fatal: destination path '.' already exists and is not an empty directory.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Build the Intel(R) SGX SDK and Intel(R) SGX PSW Package
 
 - Download the source code and prepare the submodules and prebuilt binaries:
 ```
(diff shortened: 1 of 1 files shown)


lzha101:
Please update the commit message with a Signed-off-by. 
https://github.com/intel/linux-sgx/blob/master/CONTRIBUTING.md#sign-your-work

Zildj1an:
Thanks @lzha101 commit signed

Issue #595, opened 2020-09-10
Bad libusgx_switchless.a in SGX 2.11 prebuilt SDK package


dingelish:
```
/usr/bin/ld: /opt/sgxsdk/lib64/libsgx_uswitchless.a(sl_uswitchless_untrusted.o): unable to initialize decompress status for section .debug_info
/usr/bin/ld: /opt/sgxsdk/lib64/libsgx_uswitchless.a(sl_uswitchless_untrusted.o): unable to initialize decompress status for section .debug_info
/opt/sgxsdk/lib64/libsgx_uswitchless.a: member /opt/sgxsdk/lib64/libsgx_uswitchless.a(sl_uswitchless_untrusted.o) in archive is not an object
collect2: error: ld returned 1 exit status
```

to reproduce, please `docker build` the following Dockerfiles. they all result in the above error message

Ubuntu 18.04:
```Dockerfile
FROM ubuntu:18.04
RUN apt update && apt install -y build-essential curl
RUN cd /root && \
    curl -o sdk.bin https://download.01.org/intel-sgx/sgx-linux/2.11/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.11.100.2.bin && \
    chmod +x sdk.bin && \
    echo -e 'no\n/opt' | ./sdk.bin

RUN bash -c "source /opt/sgxsdk/environment && cd /opt/sgxsdk/SampleCode/Switchless && make"
```

Centos 8
```Dockerfile
FROM centos:8

RUN dnf group install -y 'Development Tools'

RUN cd /root && \
    curl -o sdk.bin https://download.01.org/intel-sgx/sgx-linux/2.11/distro/centos8.1-server/sgx_linux_x64_sdk_2.11.100.2.bin && \
    chmod +x sdk.bin && \
    echo -e 'no\n/opt' | ./sdk.bin

RUN bash -c "source /opt/sgxsdk/environment && cd /opt/sgxsdk/SampleCode/Switchless && make"
```

lzha101:
Generally, this error should be related to different binutils used.  The SDK installer on download.01.org are built with the mitigation LVI binutils. Could you try to configure the LVI binutils in the Docker as well?

dingelish:
interesting. i can confirm use the prebuilt bintuils can fix this issue. one more question: does this means that switchless feature **_requires_** the latest binutils?

lzha101:
[...]

PR #596, opened 2020-09-11
uswitchless build: delete -g compile option

volcano0dr:
related issue: #595
Signed-off-by: volcano <volcano_dr@163.com>


yanxue-intel:
Thanks @volcano0dr , "-g" complie option should be remove on release version. But about issue:  #595,  It is not recommended to use different Binutils for SDK and Sample.

PR #602, opened 2020-09-23
Support longjmp in exception handler

guzongmin:
If the customer exception handler use longjmp function, then the memory free would be skipped. So the malloc should be changed to alloca.

### DIFF
--- a/sdk/trts/trts_veh.cpp
+++ b/sdk/trts/trts_veh.cpp
@@ -224,7 +224,8 @@ extern "C" __attribute__((regparm(1))) void internal_handle_exception(sgx_except
         continue_execution(info);
     }
 
-    if ((nhead = (uintptr_t *)malloc(size)) == NULL)
+    // The customer handler may never return, use alloca instead of malloc 
+    if ((nhead = (uintptr_t *)alloca(size)) == NULL)
     {
         sgx_spin_unlock(&g_handler_lock);
         goto failed_end;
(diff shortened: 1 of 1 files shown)


lzha101:
Thanks for raising this issue and the fix. We also need to consider the case that the alloca() triggers a stack overrun exception, although the allocated buffer is just used to save the handler pointers.

guzongmin:
> Thanks for raising this issue and the fix. We also need to consider the case that the alloca() triggers a stack overrun exception, although the allocated buffer is just used to save the handler pointers.

I think it is the second phrase execption handling, so even if the alloca function trigger #PF we still be able to handle it within another exception handling flow. 

[original exception handler (phrase 1)]
                    |
 [original exception handler (phrase 2)]  --------------> [#PF handler (phrase 1)] 
                                                                                                           |
 [original exception handler (phrase 2)] <----------------------- (ERESUME)
                    |
 [original exception handler (user phrase)]

yuyuany:
I agree the exception can be handled, and SGX_ERROR_STACK_OVERRUN is returned if stack cannot grow. But we do not want exception happens in trts. It may make users confused.

guzongmin:
> I agree the exception can be handled, and SGX_ERROR_STACK_OVERRUN is returned if stack cannot grow. But we do not want exception happens in trts. It may make users confused.

I think the fix is exactlly same as tRTS using a function with large local variables. BTW: If we do not use alloca, how can we fix this bug?

PR #603, opened 2020-09-23
Fix links to downloads for linux-latest

sbellem:
The current link (https://download.01.org/intel-sgx/latest/intel-latest/) leads to a 404.

This tiny fix may also be applied to the `sgx_2.10_reproducible` branch.


linux/reproducibility/ae_reproducibility_verifier/README.md
```diff
@@ -17,7 +17,7 @@ So this doc will give an approach using your local AE builds to verify Intel(R)
 * <a id="materials">Prepare below 4 materials</a>
 
  1. intel_signed_ae:  
- It is the Intel(R) prebuilt AE to be verified, and can be downloaded from (https://download.01.org/intel-sgx/latest/intel-latest/)
```
lzha101, line 20: We have two prebuilt AE packages. One is placed under https://download.01.org/intel-sgx/latest/dcap-latest/linux/ and the other is under https://download.01.org/intel-sgx/latest/linux-latest/. So using linux-latest may not be that correct.  Could you help to add the other place as well?

sbellem: @lzha101 just let me know if that is ok: 1f07408cade1797f0f72b7cd23252e07b0cab6e4

PR #621, opened 2020-10-21
Enhance SGX simulation mode to support TLS variables

guzongmin:
The current simulation mode is using a hack to support TLS.
In this patch, prctl syscall is used to change the FS/GS base.


lzha101:
This enhancement has been applied into the previous SGX 2.13 release. Suppose we can close this PR now.

henrywang8atfbdotcom:
We noticed that this was reverted in the upgrade to 2.15. Is there a reason for this?

lzha101:
> We noticed that this was reverted in the upgrade to 2.15. Is there a reason for this?

It introduces some new issues to work with simulation exception handling support. So we decided to revert this PR in later releases.

PR #635, opened 2020-11-23
Fix path for download_prebuilt.sh

PumpkinSeed:
The `server_url_path`  has a `/` at the and, however the files add it as well, duplicated the slash. This will throw a 404 error.


lzha101:
Could you tell us your development environment info?  We never get a 404 error with current script, although the '/' is not necessary.

PumpkinSeed:
@lzha101 It's CentOS 8. This is clearly not a path how it should look like: 
`https://download.01.org/intel-sgx/sgx-linux/2.12//prebuilt_ae_2.12.tar.gz`

Issue #639, opened 2020-12-08
Update annotations in common/inc/sgx_tseal.h


yang8621:
suggested changes to aviod confusion:
 
 typedef struct _aes_gcm_data_t
 {
-    uint32_t  payload_size;                   /*  0: Size of the payload which includes both the encrypted data and the optional additional MAC text */
-    uint8_t   reserved[12];                   /*  4: Reserved bits */
+    uint32_t  payload_size;                   /*  0: Size of the payload which includes the encrypted data */
+    uint8_t   reserved[SGX_SEAL_IV_SIZE];     /*  4: Reserved bits used as iv */
     uint8_t   payload_tag[SGX_SEAL_TAG_SIZE]; /* 16: AES-GMAC of the plain text, payload, and the sizes */
-    uint8_t   payload[];                      /* 32: The payload data which includes the encrypted data followed by the optional additional MAC text */
+    uint8_t   payload[];                      /* 32: The payload data which includes the encrypted data followed by payload_tag */
 } sgx_aes_gcm_data_t;

 typedef struct _sealed_data_t
 {
     sgx_key_request_t  key_request;       /* 00: The key request used to obtain the sealing key */
-    uint32_t           plain_text_offset; /* 64: Offset within aes_data.playload to the start of the optional additional MAC text */
-    uint8_t            reserved[12];      /* 68: Reserved bits */
-    sgx_aes_gcm_data_t aes_data;          /* 80: Data structure holding the AES/GCM related data */
+    uint32_t           plain_text_offset; /* 512: Offset within aes_data.playload to the start of the optional additional MAC text */
+    uint8_t            reserved[12];      /* 516: Reserved bits */
+    sgx_aes_gcm_data_t aes_data;          /* 528: Data structure holding the AES/GCM related data */
 } sgx_sealed_data_t;

yang8621:
https://github.com/intel/linux-sgx/pull/640

PR #640, opened 2020-12-08
Update annotations of sgx_sealed_data_t and sgx_aes_gcm_data_t (#639)

yang8621:
Signed-off-by: Huang Yang <yang.huang@intel.com>

### DIFF
--- a/common/inc/sgx_tseal.h
+++ b/common/inc/sgx_tseal.h
@@ -47,18 +47,18 @@
 
 typedef struct _aes_gcm_data_t
 {
-    uint32_t  payload_size;                   /*  0: Size of the payload which includes both the encrypted data and the optional additional MAC text */
-    uint8_t   reserved[12];                   /*  4: Reserved bits */
+    uint32_t  payload_size;                   /*  0: Size of the payload which includes the encrypted data: payload[] */
+    uint8_t   reserved[SGX_SEAL_IV_SIZE];     /*  4: Reserved bits used as iv */
     uint8_t   payload_tag[SGX_SEAL_TAG_SIZE]; /* 16: AES-GMAC of the plain text, payload, and the sizes */
-    uint8_t   payload[];                      /* 32: The payload data which includes the encrypted data followed by the optional additional MAC text */
+    uint8_t   payload[];                      /* 32: The payload data which includes the encrypted data followed by payload_tag */
 } sgx_aes_gcm_data_t;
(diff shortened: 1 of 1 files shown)


common/inc/sgx_tseal.h
```diff
@@ -47,18 +47,18 @@
 
 typedef struct _aes_gcm_data_t
 {
-    uint32_t  payload_size;                   /*  0: Size of the payload which includes both the encrypted data and the optional additional MAC text */
-    uint8_t   reserved[12];                   /*  4: Reserved bits */
+    uint32_t  payload_size;                   /*  0: Size of the payload which includes the encrypted data: payload[] */
+    uint8_t   reserved[SGX_SEAL_IV_SIZE];     /*  4: Reserved bits used as iv */
```
lzha101, line 51: Although the value of the used IV in tseal library is the same as this reserved bits array, but they are indeed different buffers.  This reserved array is actually to make the structure 16byte aligned.

yang8621: See the sample code, it's used as IV:
https://github.com/intel/linux-sgx/blob/master/SampleCode/LocalAttestation/EnclaveInitiator/EnclaveMessageExchange.cpp#L199
https://github.com/intel/linux-sgx/blob/master/SampleCode/LocalAttestation/EnclaveInitiator/EnclaveMessageExchange.cpp#L207

common/inc/sgx_tseal.h
```diff
@@ -47,18 +47,18 @@
 
 typedef struct _aes_gcm_data_t
 {
-    uint32_t  payload_size;                   /*  0: Size of the payload which includes both the encrypted data and the optional additional MAC text */
-    uint8_t   reserved[12];                   /*  4: Reserved bits */
+    uint32_t  payload_size;                   /*  0: Size of the payload which includes the encrypted data: payload[] */
```
lzha101, line 50: The payload[] includes encrypted data and the optional additional authenticated data if users seal data with a plaintext (for example, `sgx_seal_data()` is called with `p_additional_MACtext` not NULL). So I suppose the previous annotation is correct.

common/inc/sgx_tseal.h
lzha101: OK. Thanks for the explanation. This usage may not be aligned with the initial design of the structure. But the functionality is correct.

PR #656, opened 2021-01-18
Create .gitignore for SampleEnclave

lengyijun:


haitaohuang:
need Signed-off-by in commit message, see CONTRIBUTING.md

lzha101:
If we want to do so, is it better to apply this to all the provided sample code?

jbdelcuv:
> If we want to do so, is it better to apply this to all the provided sample code?

I agree, we should do the same for all sample code.
Does it make more sense updating the .gitignore file in the root of the repo?

PR #691, opened 2021-03-24
Fix cmake3 version check

jiazhang0:
If the minor version of cmake is larger than 10 but lower
than 20, e.g, 3.17.2, the check logic will think the
installed cmake is lower than cmake 3.2.x.

Signed-off-by: Jia Zhang <zhang.jia@linux.alibaba.com>


jiazhang0:
@llly Please help to see this PR. Thanks!

llly:
Your change fixes a scenario that 2.10 is higher than 3.2. Since highest minor version before 3.X is 8(0.8, 1.8, 2.8), current checking doesn't cause a failure.
The change is reasonable.

guzongmin:
@llly The compiling environment on Aliyun is fixed to some special versions, so this fix is very important for the users who need to rebuild SGX SDK and PSW on Aliyun. Can you please help to merge it after testing?

PR #725, opened 2021-07-09
docker: run aesm with root

haitaohuang:
Groups/permissions in container do not always match with host env.
Run aesm with root to ensure access to /dev/sgx_provision


haitaohuang:
This is also reported in https://github.com/intel/linux-sgx/pull/583

PR #730, opened 2021-07-21
Replace /bin/cp with cp for nix-build

sbellem:
When using nix-build, /bin/cp cannot be found as there's nothing under
/bin except for sh.

Signed-off-by: Sylvain Bellemare <sbellem@gmail.com>

Unless there's a good reason to keep `/bin/cp`, it seems that simply using `cp` is preferable as suggested in https://discourse.nixos.org/t/bin-cp-cannot-be-found-when-using-nix-build/13683/2.

### DIFF
--- a/buildenv.mk
+++ b/buildenv.mk
@@ -70,7 +70,7 @@ LINUX_UNITTESTS       := $(ROOT_DIR)/unittests
 DCAP_DIR              := $(LINUX_EXTERNAL_DIR)/dcap_source
 LIBUNWIND_DIR         := $(ROOT_DIR)/sdk/cpprt/linux/libunwind
 
-CP    := /bin/cp -f
+CP    := cp -f
 MKDIR := mkdir -p
 STRIP := strip
 OBJCOPY := objcopy
--- a/psw/ae/aesm_service/source/CMakeLists.txt
+++ b/psw/ae/aesm_service/source/CMakeLists.txt
@@ -39,7 +39,7 @@ set(CMAKE_SKIP_BUILD_RPATH true)
 ########## SGX SDK Settings ##########
 set(SGX_ARCH x64)
 execute_process (
-    COMMAND /usr/bin/getconf LONG_BIT
+    COMMAND getconf LONG_BIT
     OUTPUT_VARIABLE LONG_BIT
 )
 


buildenv.mk
```diff
@@ -58,7 +58,8 @@ LINUX_UNITTESTS       := $(ROOT_DIR)/unittests
 DCAP_DIR              := $(LINUX_EXTERNAL_DIR)/dcap_source
 LIBUNWIND_DIR         := $(ROOT_DIR)/sdk/cpprt/linux/libunwind
 
-CP    := /bin/cp -f
+#CP    := /bin/cp -f
```
veehaitch, line 61:
```suggestion
```

I'd suggest to just delete that line instead of commenting it out. If `mkdir` works, `cp` should also work as both are usually part of [`coreutils`](https://www.gnu.org/software/coreutils/).

veehaitch (approved):
Please merge this to get one step closer to pure reproducible builds using Nix.

buildenv.mk
sbellem: Thanks! Forgot that somehow. Removed and rebased.

veehaitch:
Related https://github.com/intel/SGXDataCenterAttestationPrimitives/pull/205

veehaitch:
@sbellem would you mind adding this commit https://github.com/veehaitch/linux-sgx/commit/f3c08923a17b27b0ca99ec202f656a8aa9920113? Patching the `getconf` FHS path is also required to build the PSW.

veehaitch:
@lzha101 anything we can do to get this merged?

sbellem:
@veehaitch:
 
> would you mind adding this commit [veehaitch@f3c0892](https://github.com/veehaitch/linux-sgx/commit/f3c08923a17b27b0ca99ec202f656a8aa9920113)? Patching the `getconf` FHS path is also required to build the PSW.

Added.

lzha101:
> When using nix-build, /bin/cp cannot be found as there's nothing under
/bin except for sh.

Note that we are using Nix instead of NixOS for reproducible build of several components only. So SGX code doesn't have such problem.

veehaitch:
Thanks for merging this PR! 🙏🏻

> Note that we are using Nix instead of NixOS for reproducible build of several components only. So SGX code doesn't have such problem.

Please note that this change helps you as much as anybody else using Nix to build your software and is not related to NixOS. In fact, it is necessary to have proper reproducibility. You just didn't run into problems with regard to absolute paths to binaries as you have been using Nix on an Ubuntu host _with sandboxing disabled_:

https://github.com/intel/linux-sgx/blob/e0cce77e910c4b265681597da4109f5e40f950b4/linux/reproducibility/Dockerfile#L7

From the `sandbox` section of [_nix.conf(5)_](https://manpages.debian.org/testing/nix-bin/nix.conf.5.en.html):

> If set to true, builds will be performed in a sandboxed environment, i.e., they’re isolated from the normal file system hierarchy and will only see their dependencies in the Nix store, the temporary build directory, private versions of /proc, /dev, /dev/shm and /dev/pts (on Linux), and the paths configured with the sandbox-paths option. This is useful to prevent undeclared dependencies on files in directories such as /usr/bin. In addition, on Linux, builds run in private PID, mount, network, IPC and UTS namespaces to isolate them from other processes in the system (except that fixed-output derivations do not run in private network namespace to ensure they can access the network).

Sandboxing is enabled by default. Disabling it means your Nix builds are not hermetic and might contain additional impurities. Against this background, avoiding absolute FHS paths is even a prerequisite for having proper reproducible builds using Nix.

PR #735, opened 2021-07-27
Fix aesm_service optimization issue for GCC11

### LINKED ISSUE #733: SGX 2.14 AESM service cannot find the production signature for the LE enclave
This issue is referring to https://github.com/intel/linux-sgx/blob/5178e0c26b017a05ec41f6c8edf3cbe30736e2ee/psw/ae/aesm_service/source/oal/linux/aesm_util.cpp#L56

In PSW @ `v2.14`, the resolution of the signature file for LE fails, even though `le_prd_css.bin` is installed at `<installdir>/aesm` directory, while AESM service is trying to pick it up at `<installdir>/aesm/bundles` directory.

The reason is because possibly the LE loader code is moved into `lible_launch_service_bundle.so` located at `<installdir>/aesm/bundles` directory. Therefore, `aesm_get_path` resolves `AESM_LOCATION_EXE_FOLDER` constants to the `bundles` folder, instead of the directory in which the AESM executable actually resides, because this information is inferred from the dynamic loader.

A temporary fix on the user side is to move the signature BIN file into `bundles`, so that at least AESM can launch. However, the same issue may still stop other architecture enclaves from loading, including QE, PVE, PCE and so on.

Maybe we need a more declarative and robust way to infer the AESM service directory? Note that the SystemD service unit does designate an environment variable `AESM_PATH`, which seems much more authoritative.

llly:
When compiling by GCC 11, tail call optimization breaks `__builtin_return_address` in aesm_service.

When calling `__builtin_return_address (0)` in liboal.so
https://github.com/intel/linux-sgx/blob/5178e0c26b017a05ec41f6c8edf3cbe30736e2ee/psw/ae/aesm_service/source/oal/linux/aesm_util.cpp#L51
backtrace should be:
```
0:aesm_get_path() liboal.so
1:aesm_get_pathname() liboal.so
2:load_enclave_only() lible_launch_service_bundle.so
```
`__builtin_return_address (0)` returns address of `aesm_get_pathname` which is in `liboal.so`.

During tail call optimization,  `aesm_get_pathname` frame is replaced by `aesm_get_path` frame.
https://github.com/intel/linux-sgx/blob/5178e0c26b017a05ec41f6c8edf3cbe30736e2ee/psw/ae/aesm_service/source/oal/linux/aesm_util.cpp#L234
backtrace becomes:
```
0:aesm_get_path() liboal.so
1:load_enclave_only() lible_launch_service_bundle.so
```

`__builtin_return_address (0)` returns address of `load_enclave_only` which is in `llible_launch_service_bundle.so` and then `dladdr` gets wrong folder.
 
This PR add `-fno-optimize-sibling-calls` to an aesm_service source file `aesm_util.cpp` to disable tail call optimization.

fixes #733

### DIFF
--- a/psw/ae/aesm_service/source/oal/CMakeLists.txt
+++ b/psw/ae/aesm_service/source/oal/CMakeLists.txt
@@ -25,5 +25,7 @@ target_compile_definitions(oal PRIVATE
 	$<$<CONFIG:Debug>:DBG_LOG>
 )
 
+set_source_files_properties(linux/aesm_util.cpp PROPERTIES COMPILE_FLAGS "-fno-optimize-sibling-calls")
+
 target_link_libraries(oal Threads::Threads ${CMAKE_DL_LIBS})
 set_property(TARGET oal APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-z,defs")


lzha101 (approved):

dingxiangfei2009:
There is an obvious fix without resorting to adding a flag to the compiler.

Refer to this code: https://github.com/intel/linux-sgx/blob/5178e0c26b017a05ec41f6c8edf3cbe30736e2ee/psw/ae/aesm_service/source/oal/linux/aesm_util.cpp#L51

Instead of `__builtin_return_address`, use `(void *) aesm_get_path` as the first argument to `dladdr`.

llly:
`aesm_get_path` was designed to get path of caller which can be in different folder, although it's now static and only used in the same library. Better to keep current implementation.

dingxiangfei2009:
Do you mean that `aesm_get_path` was designed to be a public API method? Why not just drop the `static` and make it public?

llly:
No need to change source code since ArchLinux is not officially supported.

PR #736, opened 2021-07-27
Fix aesmd service permission in linksgx.sh.

yuguorui:
Currently sgx-aesm-service does not depend on any other sgx packages, which makes it difficult to guarantee the installation order of sgx-aesm-service. Specifically, sgx-aesm-service may be installed earlier than libsgx-ae-pce, but the user group `sgx_prv` was created by the latter.

This PR ensures that no matter what the order of package installation is, aesmd can have access to the sgx_provision device file.

### DIFF
--- a/linux/installer/common/sgx-aesm-service/linksgx.sh
+++ b/linux/installer/common/sgx-aesm-service/linksgx.sh
@@ -35,9 +35,21 @@ if test $(id -u) -ne 0; then
     exit 1
 fi
 
-/usr/bin/getent group sgx_prv &> /dev/null
-if [ $? == "0" ]; then
-    /usr/sbin/usermod -aG sgx_prv aesmd &> /dev/null
+if [ -c /dev/sgx_provision -o -c /dev/sgx/provision ]; then
+    getent group sgx_prv &> /dev/null
+    if [ "$?" != "0" ]; then
+        # Add sgx_prv for dcap driver, which ensures that no matter what 
+        # the order of package installation, aesmd can have access to 
+        # the sgx_provision device file.
+        groupadd sgx_prv
+
+        if ! which udevadm &> /dev/null; then
+            exit 0
+        fi
+        udevadm control --reload || :
+        udevadm trigger || :
+    fi
+    usermod -aG sgx_prv aesmd &> /dev/null
 fi
 
 echo


fqiu1:
It's legitimate to install sgx-aesm-service without creating sgx_prv group. For instance, some users might just want to load normal enclaves and don't want to use attestation.

yuguorui:
> It's legitimate to install sgx-aesm-service without creating sgx_prv group. For instance, some users might just want to load normal enclaves and don't want to use attestation.

Yes, you are right in non-DCAP platform.
But for DCAP scenarios, which runs upon CPUs support FLC, I do not find a situation where aesmd-service needs to be installed without attestation.

linux/installer/common/sgx-aesm-service/linksgx.sh
```diff
@@ -35,8 +35,20 @@ if test $(id -u) -ne 0; then
     exit 1
 fi
 
-/usr/bin/getent group sgx_prv &> /dev/null
-if [ $? == "0" ]; then
+if [ -c /dev/sgx_provision -o -c /dev/sgx/provision ]; then
+    /usr/bin/getent group sgx_prv &> /dev/null
```
veehaitch, line 39:
```suggestion
    getent group sgx_prv &> /dev/null
```

linux/installer/common/sgx-aesm-service/linksgx.sh
```diff
@@ -35,8 +35,20 @@ if test $(id -u) -ne 0; then
...
+    if [ "$?" != "0" ]; then
+        # Add sgx_prv for dcap driver, which ensures that no matter what 
+        # the order of package installation, aesmd can have access to 
+        # the sgx_provision device file.
+        /usr/sbin/groupadd sgx_prv
```
veehaitch, line 44:
```suggestion
        groupadd sgx_prv
```

linux/installer/common/sgx-aesm-service/linksgx.sh
```diff
@@ -35,8 +35,20 @@ if test $(id -u) -ne 0; then
...
+
+        if ! which udevadm &> /dev/null; then
+            return 0
+        fi
+        udevadm control --reload || :
+        udevadm trigger || :
+    fi
     /usr/sbin/usermod -aG sgx_prv aesmd &> /dev/null
```
veehaitch, line 52:
```suggestion
    usermod -aG sgx_prv aesmd &> /dev/null
```

veehaitch (changes requested):
Please avoid absolute paths to executables. This assumes the host system adheres to the Filesystem Hierarchy Standard which is not always the case (a notable exception is Nix). Also see #730.

yuguorui:
> Please avoid absolute paths to executables. This assumes the host system adheres to the Filesystem Hierarchy Standard which is not always the case (a notable exception is Nix). Also see #730.

Sure, thanks for your comments. 

I will push a new version later.

linux/installer/common/sgx-aesm-service/linksgx.sh
```diff
@@ -35,8 +35,20 @@ if test $(id -u) -ne 0; then
...
+if [ -c /dev/sgx_provision -o -c /dev/sgx/provision ]; then
+    /usr/bin/getent group sgx_prv &> /dev/null
+    if [ "$?" != "0" ]; then
+        # Add sgx_prv for dcap driver, which ensures that no matter what 
+        # the order of package installation, aesmd can have access to 
+        # the sgx_provision device file.
+        /usr/sbin/groupadd sgx_prv
+
+        if ! which udevadm &> /dev/null; then
+            return 0
```
lzha101, line 47: Better to use `exit 0` instead of `return 0`.

yuguorui: Copy that.

yuguorui: Fixed.

PR #745, opened 2021-09-02
Ignore files generated when building SampleCode

LuminousXLB:
Related to #656


LuminousXLB:
Do you think it's better to move this block into `/SampleCode/.gitignore`?

PR #751, opened 2021-10-21
EDMM user space support for kernel interfaces

haitaohuang:
This implementation includes:
1)	the reusable Enclave Memory Manager as proposed previously on OE PR: https://github.com/openenclave/openenclave/pull/3991
2)	modifications for the Enclave Common Loader to support both the new kernel EDMM interfaces and legacy out-of-tree driver
3)	integration and tests with Intel SDK and PSW runtime support.  
Please refer to sdk/emm/README.md for a quick guide on build and test as well as a summary of current limitations:

### DIFF
--- a/sdk/emm/README.md
+++ b/sdk/emm/README.md
@@ -4,7 +4,7 @@ This directory contains an implementation of the Enclave Memory Manager proposed
 
 The instructions here are for developing and testing the EMM functionality only. Consult the main README for general usages.
 
-**Note:** This implementation is based on the current Linux kernel implementation posted [here](https://github.com/rchatre/linux/tree/sgx/sgx2_not_submitted_v1), which has not been finalized and upstreamed. As the kernel interfaces evolve, the EMM implementation and/or interface may change.
+**Note:** This implementation is based on the current Linux kernel implementation posted [here](https://github.com/rchatre/linux/tree/sgx/sgx2_submitted_v1_plus_rwx), which has not been finalized and upstreamed. As the kernel interfaces evolve, the EMM implementation and/or interface may change.
 
 Prerequisites
 -------------------------------
@@ -16,8 +16,9 @@ On Ubuntu 18.04/20.04, follow the general instructions from [here](https://wiki.
 ```
 $ git clone https://github.com/rchatre/linux.git
 $ cd linux
-$ git checkout sgx/sgx2_not_submitted_v1
+$ git checkout sgx/sgx2_submitted_v1_plus_rwx
 ```
+
 - For step 6, modify .config to set "CONFIG_X86_SGX=y".
 
 #### Verify kernel build and EDMM support


sdk/emm/ema.c
```diff
@@ -0,0 +1,1177 @@
...
+    return ret;
+}
+
+int ema_do_dealloc(ema_t *node, size_t start, size_t end)
+{
+    int prot = node->si_flags & SGX_EMA_PROT_MASK;
+    int type = node->si_flags & SGX_EMA_PAGE_TYPE_MASK;
+    int alloc_flag = node->alloc_flags & SGX_EMA_ALLOC_FLAGS_MASK;
+
+    if (alloc_flag & SGX_EMA_RESERVE)
```
johnvenn, line 761: just a question, so here, if alloc flags set to RESERVE, dealloc the whole ema node regardless of size?

johnvenn:
Another concern is about the dealloc function, in this case:
alloced area 4pages, then mistakenly dealloc 2 pages, the ema node will be split, but we don't know the new addr and size of the new node. looks like user need to track what is allocated/deallocated when the enclave is running. This might be a problem.

johnvenn:
per my understanding, the ema node should only exist during the enclave's lifetime, i.e., created/intiliazed on enclave initialization, and released on enclave_destroy, check the code, is there anywhere "destory_ema_root" invoked when enclave is destoryed?

sdk/emm/ema.c
```diff
@@ -0,0 +1,1177 @@
...
+    size_t index = 0;
+
+    while (node != root->guard) {
+        dump_ema_node(node, index++);
+        node = node->next;
+    }
+}
+
+#endif
+void destroy_ema_root(ema_root_t *root)
```
johnvenn, line 102: it should be invoked right before sgx_detroy_enclave?

haitaohuang: When enclave is destroyed/gone for any reason (fd to /dev/sgx_enclave is closed, host app crash/exit, etc.), the kernel will release whole range and EPC allocation. So there is really no need to cleanup memory inside enclave. Although free-up ema  may also be OK, we need be careful about the order relative to other cleanup (e.g., TCS), ema should be the very last to go.

haitaohuang:
> Another concern is about the dealloc function, in this case: alloced area 4pages, then mistakenly dealloc 2 pages, the ema node will be split, but we don't know the new addr and size of the new node. looks like user need to track what is allocated/deallocated when the enclave is running. This might be a problem.

It's similar to how mmap works, if user does not track and remove, EMAs(like VMAs) will be kept until enclave dies (like process dies).

sdk/emm/ema.c
haitaohuang: strictly speaking we should only dealloc [start,end) range. And this is a TODO

sdk/emm/ema.c
johnvenn: ok , i understand. thanks, haitao.

psw/urts/linux/isgx_user.h
```diff
@@ -109,11 +109,11 @@ enum sgx_page_flags {
...
 	_IOW(SGX_MAGIC, 0x03, struct sgx_enclave_set_attribute)
 #define SGX_IOC_PAGE_MODP \
-	_IOWR(SGX_MAGIC, 0x04, struct sgx_page_modp)
+	_IOWR(SGX_MAGIC, 0x05, struct sgx_page_modp)
 #define SGX_IOC_PAGE_MODT \
-	_IOWR(SGX_MAGIC, 0x05, struct sgx_page_modt)
+	_IOWR(SGX_MAGIC, 0x06, struct sgx_page_modt)
 #define SGX_IOC_PAGE_REMOVE \
-    _IOWR(SGX_MAGIC, 0x06, struct sgx_page_remove)
+    _IOWR(SGX_MAGIC, 0x07, struct sgx_page_remove)
```
haitaohuang, line 117: This change is to align current patch under review in LKML:https://patchwork.kernel.org/project/intel-sgx/cover/cover.1638381245.git.reinette.chatre@intel.com/
If you still use the branch from kernel repo mentioned in README, do NOT use the change.

Note you also need enable RWX for EAUG pages to be able to change permissions arbitrarily. (this is not supported in LKML patches.) 
Add this change for kernel after applying all EDMM patches.

 diff --git a/arch/x86/kernel/cpu/sgx/encl.c b/arch/x86/kernel/cpu/sgx/encl.c
index 37203da382f8..8f50a5c4d125 100644
--- a/arch/x86/kernel/cpu/sgx/encl.c
+++ b/arch/x86/kernel/cpu/sgx/encl.c
@@ -165,7 +165,7 @@ static vm_fault_t sgx_encl_eaug_page(struct vm_area_struct *vma,
         * TBD: Interface with user space policy to support max permissions
         * of RWX.
         */
-       prot = PROT_READ | PROT_WRITE;
+       prot = PROT_READ | PROT_WRITE | PROT_EXEC;
        encl_page->vm_run_prot_bits = calc_vm_prot_bits(prot, 0);
        encl_page->vm_max_prot_bits = encl_page->vm_run_prot_bits;

haitaohuang: LKML published patches +above RWX temporary patch are hosted here now:  https://github.com/rchatre/linux/commits/sgx/sgx2_submitted_v1_plus_rwx

Updated README

PR #773, opened 2022-01-11
Fix Nix detection

veehaitch:
Detect Nix by probing for the presence of the `NIX_STORE` environment
variable instead of `NIX_PATH`. The latter is only set in a `nix-shell`
session but isn't when building a derivation through `nix-build`. In
contrast, the `NIX_STORE` environment variable is set in both cases.

Companion PR to intel/intel-sgx-ssl#111

Signed-off-by: Vincent Haupert <mail@vincent-haupert.de>

### DIFF
--- a/buildenv.mk
+++ b/buildenv.mk
@@ -208,7 +208,7 @@ else ifeq ($(MITIGATION-CVE-2020-0551), CF)
     MITIGATION_LIB_PATH := cve_2020_0551_cf
 endif
 
-ifneq ($(origin NIX_PATH), environment)
+ifneq ($(origin NIX_STORE), environment)
 BINUTILS_DIR ?= /usr/local/bin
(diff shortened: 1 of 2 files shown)


veehaitch:
Following a discussion in intel/intel-sgx-ssl#111 with @jinghe-INTC, we should merge this PR first.

A short demonstration of `$NIX_PATH` VS. `$NIX_STORE`:

Within a `nix-shell` environment (the way this repository uses Nix):

```ShellSession
# nix-shell -p gcc --pure --command "env | grep -E 'NIX_PATH|NIX_STORE'"
NIX_STORE=/nix/store
```

Within a `nix-build` (a sandboxed build environment; preferred approach):

```ShellSession
# cat default.nix
with import <nixpkgs> { };
runCommand "test-env" { } ''
  env | grep -E 'NIX_PATH|NIX_STORE'
  mkdir $out
''
# nix-build default.nix
NIX_STORE=/nix/store
```

Please let me know if there's anything I can do to advance this PR, @lzha101. Happy to help!

PR #784, opened 2022-01-22
Fix libuwind target with pthreads

jonringer:
To fix:
```
In file included from ../include/libunwind_i.h:52,
                 from os-linux.c:29:
../include/pthread.h:37:10: fatal error: pthread_compat.h: No such file or directory
   37 | #include "pthread_compat.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/sdk/cpprt/linux/libunwind/src/Makefile.am
+++ b/sdk/cpprt/linux/libunwind/src/Makefile.am
@@ -689,6 +689,10 @@ AM_CPPFLAGS+= -I../../../../../common/inc/internal
 AM_CCASFLAGS = $(AM_CPPFLAGS)
 noinst_HEADERS += unwind/unwind-internal.h
 
+if OS_LINUX
+AM_CPPFLAGS+= -I../../../../../sdk/cpprt/linux/
+endif
+
 EXTRA_DIST =	$(libunwind_la_SOURCES_aarch64)			\
 		$(libunwind_la_SOURCES_arm)			\
 		$(libunwind_la_SOURCES_hppa)			\


veehaitch:
Thanks 🙏🏻

I believe Intel is going to insist on the `Signed-off-by` trailer at the end of the commit log message.

lzha101:
One question, are you parallelly building the sdk via `make -jN`?  In the cpprt Makefile, https://github.com/intel/linux-sgx/blob/2ee53db4e8fd25437a817612d3bcb94b66a28373/sdk/cpprt/Makefile#L75-L81

You can see the `pthread_compat.h` will be copied to `sdk/cpprt/linux/libunwind/include`.  If you are building the SDK with `make -j`, we can just change the upper Makefile to make sure the order that the copy action is before libunwind build.  
I guess we not adding `sdk/cpprt/linux` to the libunwind Makefile is to avoid wrong header file included by mistake as you can see there are some header files in both `sdk/cpprt/linux` and `sdk/cpprt/linux/libunwind/include`, such as unwind.h.

jonringer:
>  If you are building the SDK with make -j

Yes, I'm doing `make -j32 -l32`

lzha101:
@jonringer Can you try to change the Makefile https://github.com/intel/linux-sgx/blob/2ee53db4e8fd25437a817612d3bcb94b66a28373/sdk/cpprt/Makefile#L78-L80 to: 
````
$(CPPRT): $(OBJS) prepare-libunwind
        $(Make) libunwind
	$(CP) $(LIBUNWIND_DIR)/src/.libs/libunwind.a $(CPPRT)
	$(AR) rs $@ $(OBJS)
````
And see if the above change could fix your problem?

PR #788, opened 2022-01-27
Add support for recursive mutex in pthread lib

henrywang8atfbdotcom:
Please let me know if there's anything else I need to do to support this change.

Thanks,
Henry


lzha101:
Thanks for the patch. We are considering an overall design for the pthread attribute support. So we intend not to accept the PR right now.

PR #789, opened 2022-01-27
Do traditional Ocall if g_uswitchless_handle is uninitialized

henrywang8atfbdotcom:
Global object initialization (init_global_object) happens before
g_uswitchless_handle is initialized (via sl_init_switchless).
Some ctors invoke syscalls via switchless ocalls. This causes
init_tswitchless_ocall_mngr to be invoked with sl_call_once
prematurely, returning -1. Subsequent invokations will always return -1, such
that switchless ocalls never happen. A simple solution is to fallback
to traditional ocalls until g_uswitchless_handle has been
initialized.


lzha101 (approved):
LGTM.

PR #790, opened 2022-01-27
Add support for pthread_cond_timedwait

henrywang8atfbdotcom:


psw/urts/enclave_mutex.cpp
```diff
@@ -93,3 +129,19 @@ extern "C" int sgx_thread_setwait_untrusted_events_ocall(const void *waiter, con
...
+
+extern "C" int sgx_thread_settimedwait_untrusted_events_ocall(const void *waiter, const void *self, unsigned long long dl_sec, long dl_nsec)
+{
+    int ret = sgx_thread_set_untrusted_event_ocall(waiter);
+    if (ret != SGX_SUCCESS) return ret;
+
+    return sgx_thread_timedwait_untrusted_event_ocall(self, dl_sec, dl_nsec);
+}
+
+extern "C" int sgx_thread_settimedwait_untrusted_events_ocall(const void *waiter, const void *self, unsigned long long dl_sec, long dl_nsec)
```
henrywang8atfbdotcom, line 141: Duplicated.

lzha101:
Thanks for submitting the PR.  In this PR you want to add a pthread condition function with an absolute time timeout.  
Is there any reason/business requirement that you want this API?

The reason we didn't add this API in SDK is for security consideration.  As this API depends on the system time which is not trust.  And as you know, we don't support the time functions in trusted C library.  But we added a similar _pthread_wait_timeout() and make it internal as internal usage.

henrywang8atfbdotcom:
We just wanted to add support for pthread_cond_timedwait, which specifies abstime in its API. But now that I looks at it again, we could have just reused se_event_wait_timeout instead of introducing se_event_wait_deadline.

PR #813, opened 2022-04-05
EDMM v3

haitaohuang:
This implementation includes:

1. the reusable Enclave Memory Manager as proposed previously on OE PR: Add SGX Enclave Memory Manager openenclave/openenclave#3991
2. modifications for the Enclave Common Loader to support both the new kernel EDMM interfaces and legacy out-of-tree driver
3. integration and tests with Intel SDK and PSW runtime support.

Please refer to [README](https://github.com/intel/linux-sgx/blob/edmm_v3/sdk/emm/README.md) for a quick guide on build and test as well as a summary of current limitations.

This PR is based on kernel patches v3.
For kernel patches v1 and v2, please use edmm_v1 and edmm_v2 branches respectively.


psw/enclave_common/sgx_mm_ocalls.cpp
```diff
@@ -0,0 +1,407 @@
...
+    }
+    if (s_driver_type == SGX_DRIVER_OUT_OF_TREE)
+    {
+        return mprotect((void *)addr, length, PROT_WRITE | PROT_READ);
+    }
+    int enclave_fd = get_file_handle_from_address((void *)addr);
+    if (enclave_fd == -1)
+        return ret;
+    int map_flags = MAP_SHARED | MAP_FIXED;
+    //!TODO: support COMMIT_NOW when kernel supports
```
johnvenn, line 54: so kernel still doesn't support this operation?

haitaohuang:
replaced with  V4:https://github.com/intel/linux-sgx/pull/823

PR #815, opened 2022-04-07
Fixing SGX GDB not working because dyn_sec/elrange_start_address pointer is zero

gilanghamidy:
Recently I faced an issue that sgx-gdb refuses to load symbol and add breakpoints on enclave even though it has been properly compiled in hardware debug mode. The symptom are like in this GDB output

image

After further investigation to the Python script, the error message came from around this code in function `retrieve_enclave_info`

https://github.com/intel/linux-sgx/blob/edfe42a517b3e4b1d81204c3cdef6da6cb35fefc/sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py#L419-L424

Where the `td_addr` tried to load `thread_data_t` from invalid address. Looking up further, `td_addr` is computed by adding `tcs_t_tuple[7]`, which is `tcs.of_base` and `info_tuple[8]` which is `debug_enclave_info.elrange_start_address`. 

From my examination, `debug_enclave_info.elrange_start_address` has 0 value, yet the if is evaluated as false since `debug_enclave_info.elrange_start_address != debug_enclave_info.start_addr`. Meanwhile, `debug_enclave_info.start_addr` is actually store the actual start `mmap`ed address of the enclave. This results in the `td_addr` to be calculated with 0 offset, hence resulting invalid address when the `retrieve_enclave_info` tried to load the information from the `thread_data` object inside the enclave memor.

Fixing this is simply adding the if clause to check whether the `info_tuple[8]` or `elrange_start_address` equals to 0, then it automatically uses `info_tuple[1]` value, which is the `start_addr`. Below is the result from GDB

image

### DIFF
--- a/sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
+++ b/sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
@@ -416,7 +416,7 @@ def retrieve_enclave_info(info_addr = 0):
             td_fmt = '4Q'
 
         #get thread_data_t address
-        if (info_tuple[8] == info_tuple[1]) or ((info_tuple[3] & ET_SIM) == ET_SIM):
+        if (info_tuple[8] == 0) or (info_tuple[8] == info_tuple[1]) or ((info_tuple[3] & ET_SIM) == ET_SIM):
             td_addr = tcs_t_tuple[7] + info_tuple[1]     #thread_data_t = tcs.of_base + debug_enclave_info.start_addr
         else:
             td_addr = tcs_t_tuple[7] + info_tuple[8]     #thread_data_t = tcs.of_base + debug_enclave_info.elrange_start_address


JunjunGu1:
hi @gilanghamidy , can you give me the version of libsgx_urts.so and sgx-gdb?
I suppose you don't set elrange_start_address to 0, right?

gilanghamidy:
@JunjunGu1 hey there. I am using the latest source from the GitHub repo. I am using Fedora so I am not using the package repo as in Ubuntu. Also of course I am not setting the `elrange_start_address` to zero, since I am not modifying the internal SDK, and that value is from the `debugger_support.cpp` and I have no necessity to actually modify that part of the SDK.

gilanghamidy:
Now that you mentioned about URTS version, I just realized that apparently the Fedora installation provides `libsgx_urts.so` on its `/lib64/` folder, which actually the reason why SGX loads that library in the GDB output. I thought that the `sgx-gdb` shell script overrides the `LD_LIBRARY_PATH` before calling the actual GDB, so I thought it will load instead the URTS from the SDK installation folder. I am also compiling using the SDK and I thought it would link to the SDK installation folder.

I haven't check by removing the offending libsgx_urts.so from `/lib64/` folder, but I also have no idea when it is got installed to that folder. I thought Fedora does not provide any SGX library on their repo, right?

If this is the case, perhaps the loaded and used `libsgx_urts.so` is outdated. Is it possible that we actually implement the version checking in the GDB plugin so that it will produce correct error message if the GDB plugin detects an outdated URTS version? Also, is it possible to make URTS statically linked to the executable?

gilanghamidy:
Apparently incorrect installation of SGX PSW, but perhaps the SGX GDB can properly detect the installed PSW and warn user in case of mismatch.

Issue #809, opened 2022-03-11
SampleDNNL throws error "Illegal instruction"


Praneshss:
I am running Ubuntu 20.04. I was able to build and run `SampleDNNL` code using SIM and HW mode. The code runs without errors in SIM mode, but gives an error in HW mode.  I'm able to run other SampleCodes with HW mode perfectly.

```
~/linux-sgx/SampleCode/SampleDNNL$ ./app 
Intel(R) Deep Neural Network Library (DNNL)
Illegal instruction (core dumped)
~/linux-sgx/SampleCode/SampleDNNL$ 

```
How to debug this?
I didnt get much help using `sgx-gdb `as well.
~```
/linux-sgx/SampleCode/SampleDNNL$ sgx-gdb ./app 
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Source directories searched: ~/linux-sgx/linux/installer/bin/sgxsdk/lib64/gdb-sgx-plugin:$cdir:$cwd
Setting environment variable "LD_PRELOAD" to null value.
Reading symbols from ./app...
Loading libc++ pretty-printers.
(gdb) r
Starting program: ~/linux-sgx/SampleCode/SampleDNNL/app 
detect urts is loaded, initializing
Function "notify_gdb_to_update" not defined.
Function "sgx_debug_load_state_add_element" not defined.
Function "sgx_debug_unload_state_remove_element" not defined.
Function "urts_add_tcs" not defined.
Python Exception <class 'gdb.error'> No symbol "g_debug_enclave_info_list" in current context.: 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Intel(R) Deep Neural Network Library (DNNL)
[...]

PR #817, opened 2022-04-08
Update get_max_tcs_num() API for SGX1.

lzha101:
To fix issue https://github.com/intel/linux-sgx/issues/809

Signed-off-by: Zhang Lili <lili.z.zhang@intel.com>

### DIFF
--- a/common/inc/internal/global_data.h
+++ b/common/inc/internal/global_data.h
@@ -56,6 +56,7 @@ typedef struct _global_data_t
     sys_word_t     rsrv_executable;
     sys_word_t     thread_policy;
     sys_word_t     tcs_max_num;
+    sys_word_t     tcs_num;
     thread_data_t  td_template;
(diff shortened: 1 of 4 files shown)


michaelkaplan13:
Would it be possible someone to expand on the functional impact of this PR? I'm currently hitting an "Illegal Instruction" is inside a hardware enclave around when ECALLs are being handled to perform remote attestation and am wondering if this is related.

lzha101:
> Would it be possible someone to expand on the functional impact of this PR? I'm currently hitting an "Illegal Instruction" is inside a hardware enclave around when ECALLs are being handled to perform remote attestation and am wondering if this is related.

This PR is to correct the tcs_num setting for SGX OpenMP/DNNL on SGX1 platform. Otherwise, thread creation in OpenMP may be failed with OUT_OF_TCS error.  I think this is not related to remote attestation functionality.  Suggest you to use `sgx-gdb` to debug into your enclave and see what happens.

PR #822, opened 2022-04-14
[FIX] Memory leak in _pthread_tls_destructors

gilanghamidy:
`_pthread_tls_destructors` checks if the m_pthread value is not `NULL`. However, there are cases where the enclave code does not spawn a thread, but make a proper call to pthread TLS infrastructure. If so, the linked-list storing the TLS is not properly cleaned up after enclave exit.

This is the case found when using OpenSSL library, particularly when integrating the libssl into the SGX enclave to allow terminating SSL connection inside the enclave. We found the case when performing stress-testing on unit-test function of our implementation that uses OpenSSL's TLS.

**Additional background:** I developed a valgrind-like infrastructure to check memory leak inside the enclave while running in the hardware-debug mode. Below is the result of the memory allocation trace for the leaked memory

```
#0 - __morestack 
#1 - gmhmd_core_Memcheck_alloc 
#2 - __wrap_calloc 
#3 - _pthread_findstorage 
#4 - pthread_getspecific 
#5 - ossl_init_thread_start 
#6 - ERR_get_state 
#7 - ERR_clear_error 
#8 - state_machine.part 
#9 - gmhmd_core_TLSConnectionHandler 
#10 - sgx_gmhmd_core_TLSConnectionHandler 
#11 - do_ecall
```

```
#0 - __morestack 
#1 - gmhmd_core_Memcheck_alloc 
#2 - __wrap_calloc 
#3 - _pthread_findstorage 
#4 - pthread_getspecific 
#5 - OPENSSL_thread_stop 
#6 - gmhmd::TLSConnection::~TLSConnection 
#7 - gmhmd_core_RestAPISendResponse 
#8 - sgx_gmhmd_core_RestAPISendResponse 
#9 - do_ecall
```

OpenSSL uses thread local storage to store its error information, similar to `errno` in POSIX system, and to allow it to be thread-safe. Therefore, the OpenSSL library will call pthread's function to store values in thread local. Since SGX's pthread implementation does not check if a thread is created **inside** or **outside** the enclave while storing value in thread local storage, the `pthread_info_tls.m_pthread` will stay `NULL` at the destruction point where it is being checked.

https://github.com/intel/linux-sgx/blob/e1a37bdc381672509e49e34885730437366194e4/sdk/pthread/pthread_tls.cpp#L171-L178

OpenSSL supports calling `OPENSSL_thread_stop` to perform manual clean-up especially in the case where the destructor is not bound to the thread destructor, in some specific case like statically linking OpenSSL to the executable. But in here, it is not the case since the leak is actually due to the SGX pthread implementation does not properly clean up the thread local storage before exiting the enclave. Which funnily, calling `OPENSSL_thread_stop` adds additional thread-local storage since it is seeking a key that is not available yet in the linked-list.

The fix is simple, instead checking `m_pthread`, it must check `m_local_storage` for its `NULL` value on the destruction. Logically, `m_pthread` is also not involved in any part of the destructor; only the linked-list that is stored in `m_local_storage`, as well as calling the provided destructor if it is available. So, checking `m_pthread` in here is also logically incorrect since the `m_pthread` value is not synchronized with `m_local_storage`. 

By fixing this, OpenSSL also does not require to call `OPENSSL_thread_stop` to perform manual cleanup on thread/ECALL termination, and the memory leak is disappear.

### DIFF
--- a/sdk/pthread/pthread_tls.cpp
+++ b/sdk/pthread/pthread_tls.cpp
@@ -173,7 +173,7 @@ void _pthread_tls_destructors(void)
     struct sgx_pthread_storage *rs;
     int i;
 
-    if(NULL == pthread_info_tls.m_pthread)
+    if(NULL == pthread_info_tls.m_local_storage)
         //do nothing
         return;
 


lzha101:
Thanks for the pull request. Is it possible to share the test case that could trigger this memory leak?

lzha101:
Thanks for the pull request. Is it possible to share the test case that could trigger this memory leak?

henrywang8atfbdotcom:
We are seeing a similar issue. Currently, we are able to reproduce it by simply doing something like this.

## Trusted code:
```
static pthread_key_t key;

void ecall_set_val() {
  long* p = malloc(sizeof(long));
  *p = 100;
  pthread_key_create(&key, free);
  pthread_setspecific(key, p) ;
}
```

## Main code
```
int main() {
  // This causes OOM.
  for (int i =0; i < 100000; i++) {
    ecall_set_val(eid);
  }
}

```

gilanghamidy:
> We are seeing a similar issue. Currently, we are able to reproduce it by simply doing something like this.
> 
> ## Trusted code:
> ```
> static pthread_key_t key;
> 
> void ecall_set_val() {
>   long* p = malloc(sizeof(long));
>   *p = 100;
>   pthread_key_create(&key, free);
>   pthread_setspecific(key, p) ;
> }
> ```
> 
> ## Main code
> ```
> int main() {
>   // This causes OOM.
>   for (int i =0; i < 100000; i++) {
>     ecall_set_val(eid);
>   }
> }
> ```

Woah thank you very much. I also apologize that I forgot to reply because I was too busy working with something else while in our case, we patch the SDK ourselves to fix this. This is exactly the problem, but in our case, it was inside the OpenSSL library that calls the `pthread_setspecific` and use the thread local storage.

henrywang8atfbdotcom:
We are also seeing this come from openssl's initialization. Is this the only change we need to address this? No changes to openssl right?

gilanghamidy:
Yes this is the only fix and no change in the OpenSSL, because the root cause is in the SGX's pthread library implementation.
-------- Original message --------From: Henry Wang ***@***.***> Date: 5/29/22  21:37  (GMT+01:00) To: intel/linux-sgx ***@***.***> Cc: Gilang Mentari Hamidy ***@***.***>, Author ***@***.***> Subject: Re: [intel/linux-sgx] [FIX] Memory leak in _pthread_tls_destructors (PR #822) 
We are also seeing this come from openssl's initialization. Is this the only change we need to address this? No changes to openssl right?

henrywang8atfbdotcom:
This makes sense, I see the custom destructors being invoked now. Thanks for debugging this.

henrywang8atfbdotcom:
Just wondering if making this change has any other risks? Since we need to apply a patch to the internal version we maintain. Surely there must be a reason the code was skipping destruction if `m_pthread == NULL`?  Or another way to put it - is there ever a situation where this function would be a nop?

Also, instead of checking for m_local_storage, we could also just let the for loops do their nop work?

gilanghamidy:
> Just wondering if making this change has any other risks? Since we need to apply a patch to the internal version we maintain. Surely there must be a reason the code was skipping destruction if `m_pthread == NULL`? Or another way to put it - is there ever a condition where this function would want to be nop and skip pthread TLS destruction?
> 
> Also, instead of checking for m_local_storage == NULL, we could also just remove that conditional return entirely and let the for loops do their nop work..

I think it would be unsafe to remove the conditional entirely because we should not assume that the `m_local_storage` is always populated/non-null. From what I read in the source code, the value `m_local_storage` will only be initialized after the first call to any thread-local storage related function, while the function itself is always called at the ECALL exit. CMIIW. I think being verbose is better than being implicit relying on loop's NOP. :)

The main problem is that the logic initially checks for `m_pthread` to be not NULL, assuming that `m_local_storage` value is dependent to `m_pthread`, while in fact, both of them are invariant to each other since `m_local_storage` can still has a value even though the `m_pthread` is null.

For another side effect, I am also still not entirely sure why they put the logic like this. I am assuming that this is entirely logic error. But as far as I read the code, there is no direct relation between the `m_pthread` and the `m_local_storage` because `m_pthread` is only populated when the `pthread_create` is called from inside the enclave. I think it is a part to support create thread from inside enclave. So the `m_pthread` is to store its thread handle, which is populated by the special function inside the TRTS. If we just use regular ECALL, it will remain null for the entire single ECALL scope.

Meanwhile, the `m_local_storage` is populated by user code calling pthread's thread-local infrastructure. And the thing is that from this design, the thread-local storage lifetime **does not span between multiple ECALL**. It will be cleared directly after the ECALL exits. So in my opinion, we should not rely on using this and assume that the value is still available on multiple ECALL, despite calling it from the same thread. I think it is because the threading model in SGX is different than in the user process.

lzha101:
Thanks for providing the test code and the excellent discussion above.

With current implementation,  seems the `memory leak` only happens in the main ECALL. The thread created inside enclave will call the destructor when the task is finished and thread exit.  This behavior is aligned with the standard OS behavior - tls destructor is called when the thread exit. 

IMO,  whether this is a `memory leak` is depending on the enclave configuration, specifically `TCSPolicy`.  We have two mode for an SGX enclave:  one is binding mode and the other is unbinding mode.   
a) The binding mode means the TCS binds to the untrusted OS thread.  In this case, we preserve the TLS across ECALLs.  Suppose we should not call the destructor in this mode.
 b) The unbinding mode means the TCS will not be binding to the untrusted thread. After an ECALL, the TCS will recycle and the following ECALL will use arbitrary free TCS.  In this case, we will not preserve the TLS across ECALLs.  Perhaps we can think the root ECALL returning means enclave trusted thread exits. In this case, we can call the destructor.

Therefore,  could it fix your problem if we add a check on the tcs_policy to decide whether call destructor or not?

gilanghamidy:
Ah after digging deeper down the code, apparently the TRTS have defined the required ABI here:

https://github.com/intel/linux-sgx/blob/0746daab393d95c6f92bedc4bf34d0ec0e6d0a59/sdk/trts/linux/tls_support.c#L50-L55

So perhaps if the proper checking and initialization is fixed, the memory leak is fixed as well, and the behavior may be as what you explained. However, this is largely undocumented. So I suggest that it needs to be explicitly explained (the relation between TCS policy and the pthread library behavior)

henrywang8atfbdotcom:
> IMO, whether this is a memory leak is depending on the enclave configuration, specifically TCSPolicy

That memory leak example I provided happens regardless of TCS Policy. Partly because pthread_enabled currently always resets m_local_storage as Gilang pointed out.

> thread local storage should persist across multiple ECALL

The Intel documentation isn't very clear but from my experiments in https://github.com/intel/linux-sgx/issues/842, we see __thread and thread_local variables persist accordingly to TCS Policy. However, Pthread TLS currently does not respect the TCS Policy, as we can see the pthread TLS destructors happen on every exit, so it is never persisted. Can we clarify what the intended behavior of SGX pthread TLS should be.

lzha101:
Submitted a PR https://github.com/intel/linux-sgx/pull/843 just now.  I tried the test case in #842 and got the expected result. Please help to see if this PR could fix your issue in OpenSSL as well. Thanks.

gilanghamidy:
If the sample test case is working, then the OpenSSL should be fixed as well as it is the root cause of the problem.

Issue #818, opened 2022-04-11
s_enclave_elrange_map destructor is invalid


deeglaze:
In sdk/simulation/urtssim/enclave_creator_sim.cpp the declared destructor procedure is not guaranteed to run after std::map::~map due to the semantics of static objects. The value type is trivially copyable and destructible, so it doesn't need to be heap-allocated to begin with.

vitalybuka:
To clarify:
We are seeing MemorySanitizer reports with new https://clang.llvm.org/docs/MemorySanitizer.html#use-after-destruction-detection

==9184==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 in enclave_elrange_cleanup() enclave_creator_sim.cpp:83:19
    #1 in _dl_fini (ld-linux-x86-64.so.2)
    #2 in __run_exit_handlers (libc.so.6)
    #3 in exit (libc.so.6)
    #4 in __libc_start_main (libc.so.6)
    #5 in _start

  Memory was marked as uninitialized
    #0 in __sanitizer_dtor_callback msan_interceptors.cpp:940:5
    #1 in ~__tree __tree:1793:1
    #2 in std::__msan::map<void*, enclave_elrange*, std::__msan::less<void*>, std::__msan::allocator<std::__msan::pair<void* const, enclave_elrange*> > >::~map() map:1155:5
    #3 in MSanCxaAtExitWrapper(void*) msan_interceptors.cpp:1151:3
    #4 in __run_exit_handlers (libc.so.)
    #5 in exit (libc.so.6)
    #6 in __libc_start_main (libc.so.6)
    #7 in _start

lzha101:
Thanks for reporting this. We will check and fix it later.

JunjunGu1:
I have a fix for this, please help review.
https://github.com/intel/linux-sgx/pull/831

PR #831, opened 2022-05-05
fix the issue that s_enclave_elrange_map destructor is invalid

JunjunGu1:
which is reported from https://github.com/intel/linux-sgx/issues/818

Signed-off-by: junjungu <junjun.gu@intel.com>

### DIFF
--- a/psw/enclave_common/sgx_enclave_common.cpp
+++ b/psw/enclave_common/sgx_enclave_common.cpp
@@ -75,7 +75,7 @@ static std::map<void*, int> s_hfile;            //enclave file handles for drive
 static std::map<void*, size_t> s_enclave_size;
 static std::map<void*, bool> s_enclave_init;
 static std::map<void*, sgx_attributes_t> s_secs_attr;
-static std::map<void *, enclave_elrange_t*>s_enclave_elrange_map;
+static std::map<void *, enclave_elrange_t>s_enclave_elrange_map;
 
 typedef struct _mem_region_t {
     void* addr;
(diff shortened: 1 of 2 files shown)


sdk/simulation/urtssim/enclave_creator_sim.cpp
```diff
@@ -118,30 +108,22 @@ int EnclaveCreatorSim::create_enclave(secs_t *secs, sgx_enclave_id_t *enclave_id
         LockGuard lock(&s_enclave_info_mutex);
         if (s_enclave_elrange_map.count(base_address) != 0)
         {
-            enclave_elrange_t *enclave_elrange = s_enclave_elrange_map[base_address];
-            if (enclave_elrange == NULL)
-            {
-                return SGX_ERROR_UNEXPECTED;
-            }
+            enclave_elrange_t enclave_elrange = s_enclave_elrange_map[base_address];
```
deeglaze, line 111: This is a copy that will be thrown out. Do you want a mutable reference to s_enclave_elrange_map[base_address]?

sdk/simulation/urtssim/enclave_creator_sim.cpp
```diff
@@ -257,6 +239,12 @@ int EnclaveCreatorSim::destroy_enclave(sgx_enclave_id_t enclave_id, uint64_t enc
 
     if(enclave == NULL)
         return SGX_ERROR_INVALID_ENCLAVE_ID;
+
+    if(s_enclave_elrange_map.count(enclave->get_start_address()) != 0)
+    {
+        s_enclave_elrange_map.erase(enclave->get_start_address());
```
deeglaze, line 245: erase will remove the key if it exists, and noting if not. You don't need this to be guarded by an if.

sdk/simulation/urtssim/enclave_creator_sim.cpp
```diff
@@ -118,30 +108,22 @@ int EnclaveCreatorSim::create_enclave(secs_t *secs, sgx_enclave_id_t *enclave_id
         LockGuard lock(&s_enclave_info_mutex);
         if (s_enclave_elrange_map.count(base_address) != 0)
```
deeglaze, line 109: Since the values aren't heap-allocated, and s_enclave_elrange_map[base_address] will create a new entry if it didn't exist, you don't need the two cases.

psw/enclave_common/sgx_enclave_common.cpp
```diff
@@ -788,24 +778,9 @@ extern "C" void* COMM_API enclave_create_ex(
...
-                if (s_driver_type == SGX_DRIVER_IN_KERNEL)
-                {
-                    close_file(&hdevice_temp);
-                }
-                munmap(enclave_base, virtual_size);
-                return NULL;
-            }
-            memset(tmp_enclave_elrange, 0, sizeof(enclave_elrange_t));
-            if (memcpy_s(tmp_enclave_elrange, sizeof(enclave_elrange_t), enclave_elrange, sizeof(enclave_elrange_t)))
+            enclave_elrange_t tmp_enclave_elrange;
```
deeglaze, line 781: Like below, you can just directly write into the map. memcpy_s will never fail on these values, so you're not going to be left with a meaningless key.

lzha101 (approved):

PR #858, opened 2022-06-24
Fix aesmd booting error after RPM upgrade.

yuguorui:
According to the spec [1], the scriptlet %post of a new package executes
before %preun of the old package.

This will cause the startup.sh of the new package to be executed first,
and then the cleanup.sh of the old package to be executed when
sgx-aesm-service is upgraded, and the user aesmd will be deleted,
which leading aesmd booting error.

Replace %post with %posttrans to make sure the prerequisites for service
aesmd are met.

[1]. https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/

Signed-off-by: yuguorui <yuguorui@pku.edu.cn>


xiangquanliu (approved):
Looks reasonable

lzha101 (approved):

PR #874, opened 2022-08-23
Fix pthread_self for untrusted threads that Ecall

henrywang8atfbdotcom:
Only threads created inside Enclave have pthread_info_tls.m_pthread set. If some untrusted thread does Ecall and then pthread_self() within Enclave, it will return NULL. This errorneous return value ends up causing thread synchronization errors inside the Enclave.


henrywang8atfbdotcom:
Hi Lili @lzha101,

Do you mind taking a look at this. We've been working with this patch applied for awhile. Let me know if this is good to merge in or if you have any more questions.

lzha101 (approved):
LGTM

PR #882, opened 2022-09-04
Update Makefile of dnnl for unattended installs

melhindi:
The proposed change makes it easier to perform unattended installations in e.g. docker containers or VMs.
The `git am` command applies patches and attempts to commit the changes.
Yet, in order to commit git requires `user.name` and `user.email` to be configured, which are both usually not set in containers or VMs.

Based on my testing it is enough to simply apply the patch to avoid git asking for the user's name and email.

Signed-off-by: Muhammad El-Hindi muhammad.el-hindi@cs.tu-darmstadt.com

### DIFF
--- a/external/dnnl/Makefile
+++ b/external/dnnl/Makefile
@@ -82,7 +82,6 @@ else
 	cd $(DNNL_DIR) && git apply ../sgx_dnnl.patch
 endif
 	mkdir -p $(DNNL_DIR)/build
-	@echo "success" > $(DNNL_DIR)/build/apply_patch
 
 $(LIBDNNL): $(DNNL_DIR)/build
 	cd $(DNNL_DIR)/build && cmake -DCMAKE_CXX_ENCLAVE_FLAGS="$(CXX_ENCLAVE_FLAGS)" -DCMAKE_C_ENCLAVE_FLAGS="$(C_ENCLAVE_FLAGS)" $(DNNL_CONFIG) .. && $(MAKE)
--- a/external/dnnl/Makefile
+++ b/external/dnnl/Makefile
@@ -81,11 +81,11 @@ dnnl_src:
 ifeq ($(shell git rev-parse --is-inside-work-tree), true)
 	@$(RM) -r $(DNNL_DIR)/*
 	git submodule update -f --init dnnl
-	cd $(DNNL_DIR) && git am ../sgx_dnnl.patch
+	cd $(DNNL_DIR) && git apply ../sgx_dnnl.patch
 else
 	@$(RM) -r $(DNNL_DIR)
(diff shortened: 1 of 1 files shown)


external/dnnl/Makefile
```diff
@@ -81,11 +81,11 @@ dnnl_src:
 ifeq ($(shell git rev-parse --is-inside-work-tree), true)
 	@$(RM) -r $(DNNL_DIR)/*
 	git submodule update -f --init dnnl
-	cd $(DNNL_DIR) && git am ../sgx_dnnl.patch
+	cd $(DNNL_DIR) && git apply ../sgx_dnnl.patch
```
lzha101, line 84: Thanks for the patch.  We use `git am` to ensure the commit log into the branch. Then we can check the commit log to make sure the code will only be prepared once. See https://github.com/intel/linux-sgx/pull/882/files#diff-ea5f7c6f514534e8304d6c8873da437ee4ac4bd324e6bb40b9986aa39c43c961L44. 
With your current change, the check will be invalid and the `dnnl_src` will be executed every time you type `make`.  It's OK as seems the preparation doesn't consume too much time.  But in this way the check is unnecessary and we can remove this.  Or we can modify the source check to check something else to still ensure code preparation only once.  For example, check the` $(DNNL_DIR)/build` existence instead of git commit log.

melhindi: I see...sorry, I missed this
In my use case I did not require to run `make` multiple times, hence, I probably did not notice.
I think your suggestion to check for existence of the `build` directory might do the job.
I'll give it a thought and update my PR

melhindi: Hi @lzha101 ,
I thought about it and I think following your suggestion to check the existence of the `$(DNNL_DIR)/build` folder does the job.
The nice side effect is that it simplifies/cleans up the Makefile since we can use a Makefile target to check for the folder's existence.
An alternative but similar idea would be to create a dummy file `apply_patch` and check that for existence.
I added line 85 below to exemplify this alternative. If checking for the folder's existence is enough I'll remove this line.
Let me know what you think and I will be happy to adjust the PR accordingly.

lzha101 (approved):

external/dnnl/Makefile
lzha101: Looks good to me. I think checking the folder is enough.

melhindi: ok, thanks, I removed line 85 from the PR

melhindi:
@lzha101 thanks for approving the change.
A quick question out of curiosity: Do you have an estimate when the change will be merged into master?

lzha101:
> @lzha101 thanks for approving the change. A quick question out of curiosity: Do you have an estimate when the change will be merged into master?

Sorry I missed the question. Looks this is not an urgent fix.  We plan to merge this change after disclosing our upcoming release.

PR #889, opened 2022-09-19
Fix data races around g_enclave_state

meithecatte:
Before this commit, the memory access within `do_init_enclave` setting `g_enclave_state` to `ENCLAVE_INIT_DONE` was entirely unsynchronized. This could cause the compiler to reorder this access, performing it earlier than actually written. This could then allow it to make one of the calls to `memset_s` a tailcall, for example. The only thing preventing this is the difficulty of proving that the memset doesn't alias `g_enclave_state`.

Needless to say, this is a fragile state of affairs, not to mention that later accesses to memory touched by enclave initialization would be data races.

One way to fix this would be to access `g_enclave_state` with C11 atomics of ordering `acq_rel` or stronger. However, the freestanding environment of the SDK doesn't support C11 atomics. Thus we use the existing assembly wrappers, which are sufficient as achieving `acq_rel` semantics in x86 assembly doesn't require any special instructions.

To make it less likely that a similar flaw is reintroduced, we remove the `extern` declaration of `g_enclave_state` from the header files.

### DIFF
--- a/common/inc/internal/global_data.h
+++ b/common/inc/internal/global_data.h
@@ -78,7 +78,6 @@ typedef struct _global_data_t
 extern "C" {
 #endif
 extern SE_DECLSPEC_EXPORT global_data_t const volatile g_global_data;
-extern uint32_t g_enclave_state;
 extern sdk_version_t g_sdk_version;
 extern int EDMM_supported;
 extern uint8_t  __ImageBase;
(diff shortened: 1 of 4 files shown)


yuyuany:
Thank you for reporting this issue. We expect do_init_enclave is executed in one thread only. This is achieved by the function lock_enclave(), which read/write g_enclave_state atomically. So I don't think it necessary to  synchronize g_enclave_state when setting it to ENCLAVE_INIT_DONE.

mkow:
> We expect do_init_enclave is executed in one thread only.

@yuyuany: I don't think this is enough. If setting `g_enclave_state` to `ENCLAVE_INIT_DONE` is reordered, then you may issue other ecalls in parallel, racing the half-done initialization. See: https://github.com/intel/linux-sgx/blob/4887b30b519d7bfc9f5de6e6330bc0034cce3d75/sdk/trts/trts_ecall.cpp#L379-L382 - this check may pass too early.

yuyuany (approved):

lzha101 (approved):

lzha101:
The change looks good to me. Could you please rebase the PR to latest code before we merge the change?  Thanks.

meithecatte:
I've rebased the PR. While I was at it, I improved the commit message — when rereading, I realized that it didn't actually say what the issue is. Sorry for that.

alviroiskandar (approved):

PR #902, opened 2022-10-28
Enable gdb plugin to work with enclave from buffer

gilanghamidy:
- Initially, the gdb requires the enclave file to load the symbol. However, if we create the enclave from `sgx_create_enclave_from_buffer_ex`, the file information is not populated by the URTS, which prevents the gdb plugin to load the symbols
- This patch allows the gdb plugin to hook into the `_create_enclave_from_buffer_ex` function and perform necessary patching to allow the enclave not created from file to be debugged

This fixes the long-running issue of intel/confidential-computing.sgx.sdk#45

The signature of the entry function:
```
sgx_status_t _create_enclave_from_buffer_ex(const bool debug, uint8_t *base_addr, uint64_t file_size, se_file_t& file, ...);
```

The idea is that the debugger internally stops in the entry point of `_create_enclave_from_buffer_ex` check the `file` parameter if the `file.name` is `nullptr` (pointer to `file` is stored in `$ecx`). If so, we create a temporary file in the `/tmp` directory, and dump the enclave buffer on the `base_addr` (`$rsi`). Then, we replace the `file.name` value from `nullptr` into a dynamically-allocated string that contains the path to the temporary file. During the breakpoint, we call `malloc` from gdb to obtain the dynamically-allocated buffer to store the string so that it can be stored in the normal way by the URTS.

The debugger also keeps track over the temporary buffer file and malloc-ed pointer, which after the enclave is destroyed and `fini_enclave_debug` is called, it deletes the temporary buffer file and the malloc-ed pointer by calling `free`.

In this way, no changes at all are required to the internal URTS, PSW, and API compatibility. However, it may only work as long as the `_create_enclave_from_buffer_ex` is present in the symbol table. I tested it using the normal build (non-debug PSW), and it looks like the compiler retains the local symbol so it can still be captured by the breakpoint.

### DIFF
--- a/sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
+++ b/sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
@@ -33,10 +33,12 @@ from __future__ import print_function
 import gdb
 import struct
 import os.path
+import os
 from ctypes import create_string_buffer
 import load_symbol_cmd
 import sgx_emmt
 import ctypes
+import time
 import re
 
 # Calculate the bit mode of current debuggee project
@@ -56,6 +58,7 @@ KB_SIZE = 1024
 ENCLAVE_INFO_SIZE = 8 * 7 + 2 * 4
 INFO_FMT = 'QQQIIQQQQ'
 ENCLAVES_ADDR = {}
+DUMPED_ENCLAVE = {}
 
 # The following definitions should strictly align with the struct of
 # tcs_t
@@ -354,6 +357,13 @@ class enclave_info(object):
             return -1
         except:
             return -1
+        finally:
+            # Delete the dumped enclave if any
+            global DUMPED_ENCLAVE
+            if self.enclave_path in DUMPED_ENCLAVE:
+                gdb.execute("call (void)free({0})".format(DUMPED_ENCLAVE[self.enclave_path]))
+                os.remove(self.enclave_path)
+                del DUMPED_ENCLAVE[self.enclave_path]
 
     def append_tcs_list(self, tcs_addr):
         for tcs_tmp in self.tcs_addr_list:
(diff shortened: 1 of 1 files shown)


sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
```diff
@@ -686,6 +697,46 @@ def stop(self):
...
+            file_addr = gdb.parse_and_eval("$rcx")
+            file_str = read_from_memory(file_addr, 8 + 2*4)
+            file_tuple = struct.unpack_from("QII", file_str)
+
+            if file_tuple[1] == 0:
+                # If it is null, then it does not have a file. So we dump the buffer
+
+                # dump enclave to file
+                dump_name = "/tmp/enclave_dump_{0}.bin".format(time.time())
+                buffer_ptr = gdb.parse_and_eval("$rsi")
```
xxu36, line 718: The code assumes we are under 64-bit. Currently we still need to support 32-bit so we may need to add that support as well just like we do for several other internal breakpoints

sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
```diff
@@ -686,6 +697,46 @@ def stop(self):
             gdb.execute(gdb_cmd, False, True)
         return False
 
+# This breakpoint is to handle enclave creation with buffer
+class CreateBufferEnclaveBreakpoint(gdb.Breakpoint):
+    def __init__(self):
+        gdb.Breakpoint.__init__ (self, spec="_create_enclave_from_buffer_ex", internal=1)
```
xxu36, line 703: As you mentioned in PR description, we rely on the existence of this symbol `_create_enclave_from_buffer_ex`. I think this may not be a problem because if the user is actually doing the debugging, s/he needs to install the libsgx-urts-dbgsym package anyway, and the symbols will be there.

sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
```diff
@@ -686,6 +697,46 @@ def stop(self):
             gdb.execute(gdb_cmd, False, True)
         return False
 
+# This breakpoint is to handle enclave creation with buffer
+class CreateBufferEnclaveBreakpoint(gdb.Breakpoint):
```
xxu36, line 701: The general idea looks good to me, thanks for the contribution :)

xxu36 (commented):
The general idea looks good to me, thanks for the contribution :)

sdk/debugger_interface/linux/gdb-sgx-plugin/gdb_sgx_plugin.py
gilanghamidy: Hey, thank you for your review. Indeed I was thinking the same thing about the support for 32-bit as the SDK and SGX is also support 32-bit system. However, I am not sure how to test that since I am running the 64-bit system. I assume I will need to do either cross-compiling or install a 32-bit VM. Perhaps the difference is that the debugger now needs to inspect the stack as the function parameter is now spilled to the stack instead of register. But I guess we can do that. I'll see if I can help in the coming days :)

PR #933, opened 2023-01-20
enable atomic operations in std::shared_ptr

cblokh:


henrywang8atfbdotcom:
I wonder if it's better (less invasive) to undef _LIBCPP_HAS_NO_THREADS in __sgx because this method could scale to many changes across countless callsites?  It looks like SDK forks from some upstream libcxx - not sure exactly where/when the changes came from because patch notes aren't clear. 

@lzha101, what do you think?

cblokh:
undef _LIBCPP_HAS_NO_THREADS is probably the right long term solution, but it has greater effects then just fixing shared_ptr, it will break code compilation as it will enable code flows that rely on OS threading support for example look at `# include <sched.h>` in sdk/tlibcxx/include/__threading_support and `#include <linux/futex.h>` in sdk/tlibcxx/src/atomic.cpp
In the meanwhile I think we also need a short term solution for sgxsdk 2.18

henrywang8atfbdotcom:
OK still not working for me, not sure what the difference is.

cblokh:
can you send me the output of `objdump -d -M intel enclave.so | c++filt | grep -A3 "__add_shared()>:"`

or look at the output of `g++ -Wnon-virtual-dtor -std=c++17 -fstack-protector-strong -O2 -D_FORTIFY_SOURCE=2 -UDEBUG -DNDEBUG -ffunction-sections -fdata-sections -Wall -Wextra -Winit-self -Wpointer-arith -Wreturn-type -Waddress -Wsequence-point -Wformat-security -Wmissing-include-dirs -Wfloat-equal -Wundef -Wshadow -Wcast-align -Wconversion -Wredundant-decls -DITT_ARCH_IA64  -ffreestanding -nostdinc -fvisibility=hidden -fpie -fno-strict-overflow -fno-delete-null-pointer-checks -B/usr/local/bin  -nostdinc++ -I./include -I/linux-sgx/common/inc -I/linux-sgx/common/inc/stdc++ -I/linux-sgx/common/inc/tlibc -I/linux-sgx/common/inc/stdc++ -I/linux-sgx/common/inc/internal  -I/linux-sgx/sdk/trts -I/linux-sgx/sdk/trts/linux -c src/memory.cpp -E | less` and look at the code of `__libcpp_atomic_refcount_increment`

xxu36 (approved):
Thanks for reporting this issue! I reproduced it locally.
Generally the fix looks good to me.  Suggest to change each `!defined(_LIBCPP_HAS_NO_THREADS)` to `(!defined(_LIBCPP_HAS_NO_THREADS) || defined(_LIBCPP_SGX_HAS_CXX_ATOMIC))`

lzha101 (approved):
LGTM.

PR #960, opened 2023-08-07
Added a constant-time instruction decoder to supplement the default AEX-Notify mitigation.

scottconstable:
Signed-off by: Scott Constable <scott.d.constable@intel.com>

### DIFF
--- a/sdk/trts/trts_veh.cpp
+++ b/sdk/trts/trts_veh.cpp
@@ -202,7 +202,7 @@ extern "C" void restore_xregs(uint8_t *buf);
 extern "C" void constant_time_apply_sgxstep_mitigation_and_continue_execution(sgx_exception_info_t *info,
         uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_address, uintptr_t c3_byte_address);
 
-// constant time select based on given condtion
+// constant time select based on given condition
 static inline uint64_t cselect64(uint64_t pred, const uint64_t expected, uint64_t old_val, uint64_t new_val)
 {
     __asm__("cmp %3, %1\n\t"
--- a/sdk/trts/trts_veh.cpp
+++ b/sdk/trts/trts_veh.cpp
@@ -55,6 +55,7 @@
 #include "emm_private.h"
 #include "sgx_mm_rt_abstraction.h"
 #include "sgx_trts_aex.h"
+#include "ctd.h"
 
 #include "se_memcpy.h"
 typedef struct _handler_node_t
@@ -199,18 +200,30 @@ extern "C" __attribute__((regparm(1))) void continue_execution(sgx_exception_inf
 extern "C" void restore_xregs(uint8_t *buf);
 
 extern "C" void constant_time_apply_sgxstep_mitigation_and_continue_execution(sgx_exception_info_t *info,
-        uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_page, uintptr_t c3_byte_address);
+        uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_address, uintptr_t c3_byte_address);
 
+// constant time select based on given condtion
+static inline uint64_t cselect64(uint64_t pred, const uint64_t expected, uint64_t old_val, uint64_t new_val)
+{
+    __asm__("cmp %3, %1\n\t"
+            "cmove %2, %0"
+            : "+rm"(new_val)
+            : "rm"(pred), "rm"(old_val), "ri"(expected));
+    return new_val;
+}
 
 // apply the constant time mitigation handler
 static void apply_constant_time_sgxstep_mitigation_and_continue_execution(sgx_exception_info_t *info)
 {
     thread_data_t *thread_data = get_thread_data();
-    uintptr_t code_tickle_page, c3_byte_address, stack_tickle_pages, data_tickle_page,
+    int ct_result;
+    uint64_t data_address;
+    uintptr_t code_tickle_page, c3_byte_address, stack_tickle_pages, data_tickle_address,
               stack_base_page = ((thread_data->stack_base_addr & ~0xFFF) == 0) ?
                   (thread_data->stack_base_addr) - 0x1000 :
                   (thread_data->stack_base_addr & ~0xFFF),
               stack_limit_page = thread_data->stack_limit_addr & ~0xFFF;
+    int data_tickle_address_is_within_enclave;
 
     // Determine which stack pages can be tickled
     if (((uintptr_t)info & ~0xFFF) == stack_base_page) {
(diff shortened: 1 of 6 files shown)


sdk/trts/linux/ctd.c
```diff
@@ -0,0 +1,1334 @@
+/**
+ * @file dis_main.c
```
lzha101, line 2: Can we cleanup the header declaration and add the license?

sdk/trts/linux/Makefile
```diff
@@ -87,6 +87,10 @@ $(LIBTRTS): $(OBJS) $(LIBSGX_MM)
 	echo $(ASM_SRCS)
 	$(CC) $(CFLAGS) $(TCFLAGS) -c $< -o $@
 
+# The CTD requires AVX2
+ctd.o: ctd.c
+	$(CC) -mavx2 -O3 -masm=intel $(filter-out -O2,$(CFLAGS)) $(TCFLAGS) -c $< -o $@
```
lzha101, line 92: Looks it introduces a dependency to enclaves that HW must support AVX2.

sdk/trts/linux/ctd.c
scottconstable: Fixed in the new force-push

sdk/trts/linux/Makefile
scottconstable: You are correct. It is also true that all Intel processors that support SGX also support AVX2.

sdk/trts/linux/ctd.c
```diff
@@ -0,0 +1,1359 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+/**
+ * File: ctd.h
```
JunjunGu1, line 33: should be ctd.c

sdk/trts/trts_veh.cpp
```diff
@@ -251,6 +259,29 @@ static void apply_constant_time_sgxstep_mitigation_and_continue_execution(sgx_ex
         }
     }
 
+    ct_result = ct_decode(&info->cpu_context, &data_address);
+
+    data_tickle_address = stack_tickle_pages & ~0x1;
+    data_tickle_address = cselect64(ct_result, 1, data_address, data_tickle_address);
+    data_tickle_address = cselect64(ct_result, 2, data_address, data_tickle_address);
+    data_tickle_address_is_within_enclave =
+		sgx_is_within_enclave((void*) data_tickle_address, sizeof(uint8_t));
```
JunjunGu1, line 268: should be sizeof(uintptr_t)?

sdk/trts/trts_veh.cpp
```diff
@@ -199,18 +200,30 @@ extern "C" __attribute__((regparm(1))) void continue_execution(sgx_exception_inf
...
-        uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_page, uintptr_t c3_byte_address);
+        uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_address, uintptr_t c3_byte_address);
 
+// constant time select based on given condtion
+static inline uint64_t cselect64(uint64_t pred, const uint64_t expected, uint64_t old_val, uint64_t new_val)
+{
+    __asm__("cmp %3, %1\n\t"
+            "cmove %2, %0"
+            : "+rm"(new_val)
+            : "rm"(pred), "rm"(old_val), "i"(expected));
```
JunjunGu1, line 211: It has an error in debug build
trts_veh.cpp: In function ‘uint64_t cselect64(uint64_t, uint64_t, uint64_t, uint64_t)’:
trts_veh.cpp:211:56: error: asm operand 3 probably doesn’t match constraints [-Werror]
  211 |             : "rm"(pred), "rm"(old_val), "i"(expected));
      |                                                        ^
trts_veh.cpp:211:56: error: impossible constraint in ‘asm’
cc1plus: all warnings being treated as errors

sdk/trts/linux/ctd.c
scottconstable: Fixed in a force-push

sdk/trts/trts_veh.cpp
scottconstable: Fixed in a force-push. The debug build was failing because this function wasn't being inlined by the compiler, and therefore `expected` must be a register.

sdk/trts/trts_veh.cpp
scottconstable: `sizeof(uint8_t)` is correct. This check is ensuring that the 1-byte read+write in `.ct_check_write` will be safe: https://github.com/intel/linux-sgx/blob/a1eeccba5a72b3b9b342569d2cc469ece106d3e9/sdk/trts/linux/trts_mitigation.S#L325.

sdk/trts/trts_veh.cpp
```diff
@@ -251,6 +259,29 @@ static void apply_constant_time_sgxstep_mitigation_and_continue_execution(sgx_ex
         }
     }
 
+    ct_result = ct_decode(&info->cpu_context, &data_address);
```
JunjunGu1, line 262: ct_decode is missing in simulation build

scottconstable: I wasn't able to replicate this issue, but I suspect that the root cause is that ctd.c was in `sdk/trts/linux`, and the Makefile in that directory configures the linking a little differently from the rest of the trts.

I updated the PR to move ctd.c up into `sdk/trts`. Please try the build again and let me know if it still fails.

JunjunGu1: when you build Sample Enclave, use make SGX_MODE=SIM, it will report ct_decode is undefined.

lzha101: It looks better if we can remove all the AEX-Notify related binaries from simulation.  Right now, we just disable it during runtime while the AEX-Notify logic is still included in the simulation library. 
@scottconstable you can temporarily update https://github.com/intel/linux-sgx/blob/master/sdk/simulation/trtssim/linux/Makefile#L61 to add the new ctd.o in the simulation library to let the simulation mode compilation pass.  We will consider to remove all the AEX-Notify logic from simulation after your PR.

scottconstable: Thanks! I took Lili's advice and simply updated the trtssim Makefile, and I no longer see the build failing on `ct_decode` with `SGX_MODE=SIM`.

JunjunGu1 (approved):

ShundaZhang (approved):

sdk/trts/trts_veh.cpp
```diff
@@ -199,18 +200,30 @@ extern "C" __attribute__((regparm(1))) void continue_execution(sgx_exception_inf
 extern "C" void restore_xregs(uint8_t *buf);
 
 extern "C" void constant_time_apply_sgxstep_mitigation_and_continue_execution(sgx_exception_info_t *info,
-        uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_page, uintptr_t c3_byte_address);
+        uintptr_t ssa_aexnotify_addr, uintptr_t stack_tickle_pages, uintptr_t code_tickle_page, uintptr_t data_tickle_address, uintptr_t c3_byte_address);
 
+// constant time select based on given condtion
```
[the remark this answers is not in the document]
scottconstable: @lzha101 fixed in the latest commit

lzha101 (approved):

PR #971, opened 2023-09-04
Fix memory leak in sign_tool.cpp

mspiegel:
Ownership of the parser is transferred from measure_enclave() to load_enclave(). load_enclave() passes the parser by reference to the CLoader constructor. CLoader cannot be responsible for deleting the parser. The parser must be deleted at the end of load_enclave().


lzha101 (approved):
Thanks for the PR. LGTM.

PR #1012, opened 2024-02-14
feat: make builds reproducible

haraldh:
By removing the timestamps from the bundled archives, builds can be made reproducible, which helps auditing.


lzha101:
@haraldh, Thanks for the PR. May I know what the motivation for the PR is? Do you want to reproduce the AESM service? Our current reproducible targets are for the prebuilt binaries only, such as the prebuilt libs and AEs because AEs requires to be signed by Intel and users cannot build them and replace them. While users could easily replace the untrusted components with the ones built from source by themselves.

PR #1019, opened 2024-03-03
Update __config to make clangd happy

nzpznk:
The sgx-sdk uses _LIBCPP_COMPILER_GCC macro to modify the stl. However, when using clangd, the language sever defines __clang__ macro, so that if matches __clang__ first, the _LIBCPP_COMPILER_GCC will not be defined, thus causing clangd to report error.

### DIFF
--- a/sdk/tlibcxx/include/variant
+++ b/sdk/tlibcxx/include/variant
@@ -234,7 +234,7 @@ _LIBCPP_BEGIN_NAMESPACE_STD
 // TODO: GCC 5 lies about its support for C++17 (it says it supports it but it
 //       really doesn't). That breaks variant, which uses some C++17 features.
 //       Remove this once we drop support for GCC 5.
-#if _LIBCPP_STD_VER > 14 && !(defined(_LIBCPP_COMPILER_GCC) && _GNUC_VER_NEW < 6000)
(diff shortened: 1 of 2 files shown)


sdk/tlibcxx/include/variant
```diff
@@ -234,7 +234,7 @@ _LIBCPP_BEGIN_NAMESPACE_STD
 // TODO: GCC 5 lies about its support for C++17 (it says it supports it but it
 //       really doesn't). That breaks variant, which uses some C++17 features.
 //       Remove this once we drop support for GCC 5.
-#if _LIBCPP_STD_VER > 14 && !(defined(_LIBCPP_COMPILER_GCC) && _GNUC_VER_NEW < 6000)
+#if _LIBCPP_STD_VER > 14
```
haitaohuang, line 238: you commit message seems not match the changes. Can you state the problem to solve, then describe the changes accurately?
I don't see you update variant. You also removed condition for  _LIBCPP_COMPILER_GCC. The reason for the changes is not clear to me.
also note the TODO comment. It seems we should only drop the condition if we don't support GCC5.
I'm not sure if we decided to do that. Maybe others have more ideas.

PR #1050, opened 2024-09-19
fix a header issue in sgx_enclave_common.cpp

dingelish:
sgx_enclave_common.cpp [line 164](https://github.com/intel/linux-sgx/blob/c1ceb4fe146e0feb1097dee81c7e89925443e43c/psw/enclave_common/sgx_enclave_common.cpp#L164) uses `std::upper_bound`. the corresponding header file is `algorithm`. however, it's missing in sgx_enclave_common.cpp. this patch adds it and make the package compile with the latest toolchain.

### DIFF
--- a/psw/enclave_common/sgx_enclave_common.cpp
+++ b/psw/enclave_common/sgx_enclave_common.cpp
@@ -34,6 +34,7 @@
 #include <sys/types.h>
 #include <dlfcn.h>
 #include <map>
+#include <algorithm>
 #include <functional>
 #include "sgx_enclave_common.h"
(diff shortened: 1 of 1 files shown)


jbdelcuv (approved):
Looks good to me.

bgotowal:
Hi @dingelish ! Thank you for your contribution! That was a valid concern and solution but just noticed this issue has been addressed as part of https://github.com/intel/confidential-computing.sgx/commit/6ce4f4ad0ea485cf0d1210c28414872bbefa2663 in the meantime. I will go ahead and close the PR.

Thank you!

PR #1055, opened 2024-10-01
Stop throwing away make stderr output

berrange:
Calling '$(MAKE)' against a sub-directory, while redirecting the stderr output to /dev/null makes it impossible to debug build failures, as it hides all compiler error messages. Stop sending output to /dev/null, so users can figure out what's wrong when a new compiler release fails to build SGX.


haitaohuang (approved):

berrange:
Rebased to resolved conflicts with HEAD

jbdelcuv (approved):
Looks good to me.

PR #1058, opened 2024-10-02
Disable use of undefined DEF_WEAK macro in strptime.c

berrange:
The strptime.c code invokes, what is presumably an undefined macro called, DEF_WEAK. The compiler can't resolve this to a function, so being traditional C, it invents a function taking arbitrary args and returning an int.

GCC 9.5.0 used in the reproducible build warns about this problem but sdk/Makefile.source is throwing stderr from the tlibc build by sending it all to /dev/null, so the warning is never seen.

New GCC 14 has turned this warning into a fatal error, such that this bug breaks the build outside of the NixOS envirnoment.

The other use of DEF_WEAK in tlibc/time/localtime.c was surrounded by '#if 0', so do the same in strptime.c

### DIFF
--- a/sdk/tlibc/time/strptime.c
+++ b/sdk/tlibc/time/strptime.c
@@ -89,7 +89,9 @@ strptime(const char *buf, const char *fmt, struct tm *tm)
 {
 	return(_strptime(buf, fmt, tm, 1));
 }
+#if 0
 DEF_WEAK(strptime);
+#endif
 
 static char *
 _strptime(const char *buf, const char *fmt, struct tm *tm, int initialize)


pdy:
Issue I found to be closed if PR will be accepted https://github.com/intel/linux-sgx/issues/1066

bgotowal:
Thank you for the contribution! I'd like to let you know that the issue was resolved by removing the problematic DEF_WEAK  in [8e9ed53](https://github.com/intel/confidential-computing.sgx/commit/8e9ed532cc9b4dc4f86e6d2e1fef45e411892233#diff-24f8f3fafc5866417c6295b80fc3bd7176b0a690f42e42187a18749ddedf28cb). Apologies for not getting back to this PR earlier. Can be closed.

PR #1061, opened 2024-10-08
psw/ae: blank out ld-linux.so interpretor path from AEs

### LINKED ISSUE #1040: Pre-built architectural enclaves contain redundant NixOS ld-linux path
Attempting to reproduce the architectural enclaves on a non-NixOS platform fails, despite having the same toolchain versions. Debugging revealed that this is a result of the built enclaves containing a reference to the NixOS ld-linux path.

This can be easily seen by querying the INTERP section of the binaries

```
$ readelf -l libsgx_*.signed.so | grep --after 2 INTERP
  INTERP         0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
                 0x0000000000000054 0x0000000000000054  R      0x1
      [Requesting program interpreter: /nix/store/scd5n7xsn0hh0lvhhnycr9gx0h8xfzsl-glibc-2.34-210/lib/ld-linux-x86-64.so.2]
--
  INTERP         0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
                 0x0000000000000054 0x0000000000000054  R      0x1
      [Requesting program interpreter: /nix/store/scd5n7xsn0hh0lvhhnycr9gx0h8xfzsl-glibc-2.34-210/lib/ld-linux-x86-64.so.2]
--
  INTERP         0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
                 0x0000000000000054 0x0000000000000054  R      0x1
      [Requesting program interpreter: /nix/store/scd5n7xsn0hh0lvhhnycr9gx0h8xfzsl-glibc-2.34-210/lib/ld-linux-x86-64.so.2]
--
  INTERP         0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
                 0x0000000000000054 0x0000000000000054  R      0x1
      [Requesting program interpreter: /nix/store/scd5n7xsn0hh0lvhhnycr9gx0h8xfzsl-glibc-2.34-210/lib/ld-linux-x86-64.so.2]
```

IIUC, the ld-linux interpretor is irrelevant for SGX, since the SGX SDK contains native logic for loading code into the enclaves. If this is indeed the case, then I'd request that the reference to the NixOX ld-linux is removed from future pre-built enclave binaries.

Looking at the build system this could potentially be achieved by extending the 'strip' command usage to also purge the '.interp' section of the ELF binary.

```
$(STRIP) --strip-unneeded --remove-section=.comment --remove-section=.note --remove-section=.interp $(SONAME)
```

Alternatively the `-Wl,-dynamic-linker` arg could be used to set the .interp section contents to some dummy value to indicate it is redundant / unused e.g.

```
LDTFLAGS  = -L$(SGX_LIB_DIR) -Wl,--whole-archive $(TRTSLIB) -Wl,--no-whole-archive \
            -Wl,--start-group $(EXTERNAL_LIB) -Wl,--end-group -Wl,--build-id       \
            -Wl,--version-script=$(ROOT_DIR)/build-scripts/enclave.lds $(ENCLAVE_LDFLAGS) \
            -Wl,dynamic-linker,/bin/false
```

There are probably other options too - I don't mind as long as this (apparently) redundant NixOS path can be removed from future pre-built enclave binaries.

berrange:
The enclaves are getting built as ELF executables, and thus the linker will embed the current ld-linux.so path for the host OS environment in the binary:

 $ readelf -a libsgx_pce.signed.so | grep interpreter
      [Requesting program interpreter: /nix/store/xmprbk52mlcdsljz66m8yf7cf0xf36n1-glibc-2.38-44/lib/ld-linux-x86-64.so.2]

The SGX enclaves are never loaded using ld-linux.so, as SGX has custom code for loading enclaves in the required manner.

This embedded ld-linux.so path thus serves no functional purpose, while also making it harder to do a reproducible build of the enclaves outside of the NixOS environment.

This patch blanks out the NixOS interpretor path, by setting it to the empty string.

Fixes: https://github.com/intel/linux-sgx/issues/1040

Two alternatives I considered:

 * Extending the `strip` command to add `--remove-section=.interp`.  This has the undesirable side effect that 'readelf' will complain that the AE ELF file is malformed, because ELF executables are expected to have an `.interp` section
 * Changing LDTFLAGS to add `-Wl,--shared`. This tells the linker to create an ELF file that is a shared library rather than an executable, and thus avoids creating a `.interp` section in the first place. This has a bunch of changes to the overall ELF file, however, and I was not confident that the result would be functionally correct for the SGX enclave loader.

I still believe that using `-Wl,--shared` is probably the more semantically logical approach, but using `-Wl,--dynamic-loader,` is the approach that has fewer chances of unexpected behaviour changes, hence I picked the latter here.

### DIFF
--- a/psw/ae/buildenv.mk
+++ b/psw/ae/buildenv.mk
@@ -78,7 +78,7 @@ LDTFLAGS  = -L$(SGX_LIB_DIR) -Wl,--whole-archive $(TRTSLIB) -Wl,--no-whole-archi
             -Wl,--start-group $(EXTERNAL_LIB) -Wl,--end-group -Wl,--build-id       \
             -Wl,--version-script=$(ROOT_DIR)/build-scripts/enclave.lds $(ENCLAVE_LDFLAGS)
 
-LDTFLAGS += -Wl,-Map=out.map -Wl,--undefined=version -Wl,--gc-sections
+LDTFLAGS += -Wl,-Map=out.map -Wl,--undefined=version -Wl,--gc-sections -Wl,-dynamic-linker,
 
 DEFINES := -D__linux__
 


berrange:
The DCAP counterpart of this is https://github.com/intel/SGXDataCenterAttestationPrimitives/pull/435

haitaohuang (approved):

haitaohuang:
I realize we probably do not want to re-sign AEs for this change only if this affects the sigstruct (likely not but did not check)

PR #1062, opened 2024-10-09
buildenv: enable pointing enclave build to alternative glibc headers

berrange:
The enclaves must be built with headers from a specific glibc version for the sake of reproducibility. It does not actually link to glibc, merely requiring a few self contained definitions. In the NixOS build env the glibc system include directories get set through various wrapper scripts NixOS creates.

When attempting a reproducible build outside of NixOS though, we can't rely on the compiler having the matching glibc system include dirs. Instead there needs to be a way to inject "-isystem/some/path" args into the enclave compiler flags.

This commit adds a "ENCLAVE_SYSTEM_INCLUDES" make var can be set by the person triggering 'make', to provide a way to inject system include directories to the enclave build process.

### DIFF
--- a/buildenv.mk
+++ b/buildenv.mk
@@ -285,7 +285,7 @@ endif
 # When `pie' is enabled, the linker (both BFD and Gold) under Ubuntu 14.04
 # will hide all symbols from dynamic symbol table even if they are marked
 # as `global' in the LD version script.
-ENCLAVE_CFLAGS   = -ffreestanding -nostdinc -fvisibility=hidden -fpie -fno-strict-overflow -fno-delete-null-pointer-checks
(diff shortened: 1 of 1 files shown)


haitaohuang:
Actually I think the enclave include path should be clean and only pointed to SDK trusted libc headers. I'll defer to others to review

berrange:
> Actually I think the enclave include path should be clean and only pointed to SDK trusted libc headers. I'll defer to others to review

Yeah, I was a bit surprised that the build process was seemingly relying on headers from the NixOS libc install. If it can be changed to only rely on headers from the SDK that would simplify things for OS vendors trying to reproduce the builds.

PR #1064, opened 2024-10-09
psw: prefer /dev/sgx_provision & /dev/sgx_enclave

berrange:
The default behaviour for SGX support in upstream Linux is to have /dev/sgx_provision & /dev/sgx_enclave device nodes[1], rather than /dev/sgx/provision & /dev/sgx/enclave.

The code should prefer opening the current default device nodes first, with the other paths as the fallback, so the common case will be an immediate success.

[1] https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/sgx/main.c


haitaohuang (approved):
Looks Good to me

PR #1071, opened 2025-01-30
Updated aesm_service to use C++17

jazzybluesea:
Updated aesm_service to use C++17.  This was failing compilation when compiling for Azure Linux 3.0

### DIFF
--- a/psw/ae/aesm_service/source/CMakeLists.txt
+++ b/psw/ae/aesm_service/source/CMakeLists.txt
@@ -61,7 +61,7 @@ if(REF_LE)
 endif()
 
 set(CMAKE_CXX_STANDARD_REQUIRED 1)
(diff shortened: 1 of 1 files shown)


jbdelcuv:
The change looks good to me.
It should be okay but we need to verify that all distros that we currently support support this.

bgotowal:
Hi @jazzybluesea ! Thank you for the contribution and bringing the issue to our attention! Just done a cross check and noticed the change has been already done in the meantime as part of [Bump CXX Standard for AESM Service to 17 and DCAP to include -lpthread flag for GTEST](https://github.com/intel/confidential-computing.sgx/commit/bde9b196f9920f119620e54b05f758f67c7f98c0). Closing this PR.

PR #1105, opened 2026-07-24
sdk: avoid failure due to attribute(regparm) with GCC 16

berrange:
GCC >= 16 now warns about use of attribute(regparm) on x86_64 targets. The attribute was always ignored on x86_64, so only had effect on i686 and thus the warning can be safely ignored.

As a possible alternative to this, could the use of regparm be removed entirely ?

```
$ git grep regparm
external/sgx-emm/api_tests/App/App.cpp:#define fastcall __attribute__((regparm(3),noinline,visibility("default")))
psw/urts/linux/debugger_support.cpp:#define fastcall __attribute__((regparm(3)))
psw/urts/tcs.cpp:#define fastcall __attribute__((regparm(3),noinline,visibility("default")))
sdk/compiler-rt/stack_chk.c:void __attribute__ ((regparm(1))) __intel_security_check_cookie(void * cookie)
sdk/trts/linux/tls_support.c:void * __attribute__((__regparm__(1))) ___tls_get_addr(tls_index *ti)
sdk/trts/linux/trts_pic.S: * extern "C" __attribute__((regparm(1))) void continue_execution(sgx_exception_info_t *info);
sdk/trts/linux/trts_pic.S: * extern "C" __attribute__((regparm(1))) void second_phase(sgx_exception_info_t *info, 
sdk/trts/trts_veh.cpp:extern "C" __attribute__((regparm(1))) void continue_execution(sgx_exception_info_t *info);
sdk/trts/trts_veh.cpp:extern "C" __attribute__((regparm(1))) void second_phase(sgx_exception_info_t *info, 
sdk/trts/trts_veh.cpp:extern "C" __attribute__((regparm(1))) void internal_handle_exception(sgx_exception_info_t *info)
```

I'm unclear whether SGX still considers 32-bit builds a valid target ? If x86_64 is the only valid build target, then since regparam is a no-op it could be removed. This patch takes the conservative approach and leaves it present in case it is important for 32-bit still.

### DIFF
--- a/psw/urts/linux/Makefile
+++ b/psw/urts/linux/Makefile
@@ -36,10 +36,10 @@ CXXFLAGS += -DDISABLE_TRACE
 CFLAGS += -DDISABLE_TRACE
 endif
 
-CXXFLAGS += -fPIC -Werror -g
+CXXFLAGS += -fPIC -Werror -Wno-attributes -g
 CXXFLAGS += $(ADDED_INC)
 
-CFLAGS += -fPIC -Werror -g
+CFLAGS += -fPIC -Werror -Wno-attributes -g
 CFLAGS += $(ADDED_INC)
 
 VTUNE_DIR = $(LINUX_EXTERNAL_DIR)/vtune/linux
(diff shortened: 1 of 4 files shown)


mbronk-intc:
Thank you for this proposal.
Granted, `x86_64` is the only supported build target, but to keep our options open, we're currently planning to go with a conditional 64-bit only disable in the few places that need them, similar to:
```cpp
#if defined(__x86_64__)
#  define SE_REGPARM(n)
#else
#  define SE_REGPARM(n) regparm(n)
#endif
```
It should go live in the upcoming **2.30** release.

berrange:
Sounds good to me, I'll close this alternative PR.

